PR #20, opened 2026-06-19
feat: git graph toolbar RWD, reload feedback, and VSCode-style context menus

mukiwu:
## Summary

Improves the git graph (and source control) UX in several ways:

- **Responsive toolbar** — the git graph toolbar now measures its own width and, below a threshold, folds the action icons into a single overflow (`...`) menu. The branch dropdown and search stay reachable at every width, so the toolbar no longer crowds when the graph is used as a split panel.
- **Reload feedback** — refresh in both the git graph and the source control panel now spins a disabled indicator while reloading. Local git reads finish almost instantly, so a shared `withMinDuration` helper keeps the spinner visible long enough to be perceived.
- **VSCode-style context menus** — remote branch refs gained a context menu (they were read-only), and the commit menu now matches VSCode more closely.
  - Remote branch: checkout as a local tracking branch, merge into current, pull into current, delete remote branch (confirmed), copy branch name.
  - Commit: checkout, merge, rebase current branch on this commit, copy commit hash, copy commit subject.
  - Each menu label carries its underlying git command, e.g. `Pull into current branch (git pull)`.
- **Backend** — new Rust commands (`git_rebase`, `git_branch_checkout_track`, `git_pull`, `git_push_delete`), each shelling git through the existing flag-smuggling guard.
- **Polish** — context-menu font set to 13px, tighter toolbar icon spacing, and `CommitInputModal` gained an optional message body and danger confirm button so the delete confirmation reads as a question with a red action.

## Implementation notes

- Menu assembly was extracted from `GitGraphTabContent` into a pure, unit-tested module (`lib/contextMenuItems.ts`) so item ids, order, grouping, and danger flags are testable without rendering the tab.
- `withMinDuration` lives in `src/lib` since both the git graph and source control use it.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Frontend unit tests pass (Vitest)
- [x] Rust unit tests pass (`cargo test`), including an integration test for the remote-branch actions against a local bare remote
- [x] Production build passes (`npm run build`)
- [ ] Manual: narrow the git graph panel and confirm the toolbar collapses to the overflow menu
- [ ] Manual: refresh in both the git graph and source control and confirm the spinner shows
- [ ] Manual: right-click a remote branch and exercise checkout / pull / delete; right-click a commit and try rebase


mukiwu:
Addressed the Gemini review note on `withMinDuration`: the timer is now tracked and `clearTimeout`'d in a `finally` block so a rejected `work` promise no longer leaves a dangling timer. Added a regression test asserting no pending timer remains after rejection (`8610e8c`).

PR #24, opened 2026-06-19
refactor: claude progress panel review fixes (tail-read, re-render, todo key)

### LINKED ISSUE #16: Claude 進度面板：code review 小優化（讀整檔效能 / re-render / todo key）
PR #14 的 code review 找到三個非阻塞的小問題，當時為了先上線而緩著，記在這裡

## 1. 後端每次檔案變動都讀整個 transcript

`src-tauri/src/modules/claude_progress/mod.rs` 的 notify callback 每次檔案事件都 `read_to_string` 整個 transcript，再丟掉 offset 之前的內容。長 session 的 transcript 長到幾 MB 時，每次 append 都重讀整檔，是 O(N × 檔案大小)。cursor 已經存了 byte offset，可以改成 `File::open` + `seek(SeekFrom::Start(offset))` + 只讀新增的尾段。`byte_len` 也用 `read_to_string` 純粹量長度，可改成 `metadata().len()`。順帶一提 `latest_transcript` 每次事件都重掃整個目錄，可改用事件帶的路徑、或只在 Create 事件時重掃。

## 2. 重複 todo / idle 造成多餘 re-render

`progressState.ts` 的 `todo` 跟 `idle` reducer case 即使資料沒變也回傳新物件，讓 `progressStore.pushLines` 的短路失效，每次 transcript append 都重寫 sessions、re-render 面板。可在 reducer 內短路（todos 內容相同、或已經 idle 時回傳原 state）。

## 3. todo 用 array index 當 React key

`ClaudeProgressPanel.tsx` 的 todo 清單用 index 當 key，claude 重排或插入 todo 時 React 可能重用錯 DOM 節點。改用穩定 key（需要 normalizer 給 todo 一個 id，或用 content 當 key）。

三個都是 minor，不影響日常使用

mukiwu:
## Summary

Three non-blocking code-review fixes for the Claude progress panel, carried over from PR #14.

**1. Backend tail-read instead of full re-read** (`claude_progress/mod.rs`)
The notify callback re-read the entire transcript on every file event (O(N × filesize) for long sessions). Now seeks to the stored byte offset (`File::open` + `seek`) and reads only the appended tail; `byte_len` uses `metadata().len()`. Recovery preserved: a shrunk file or a non-UTF-8 tail falls back to reading from the start. `latest_transcript` only rescans the directory on Create events.

**2. Avoid redundant re-renders** (`progressState.ts`)
The `todo`/`idle` reducer cases returned a new object even when nothing changed, defeating the `pushLines` short-circuit. They now return the same state reference when the todos are unchanged / state is already idle.

**3. Stable todo React key** (`ClaudeProgressPanel.tsx`)
Todo list keyed on array index → keyed on the todo content, so reorders/inserts don't reuse the wrong DOM node.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run` — pass (incl. new reducer reference-equality tests)
- [x] `cargo test` — pass (incl. new tail-read / shrunk-file / metadata tests)
- [x] `npm run build`

Closes #16

### DIFF
--- a/src-tauri/src/modules/claude_progress/mod.rs
+++ b/src-tauri/src/modules/claude_progress/mod.rs
@@ -4,6 +4,8 @@
 //! function so it can be tested without the filesystem or the notify watcher.
 
 use std::collections::HashMap;
+use std::fs::File;
+use std::io::{Read, Seek, SeekFrom};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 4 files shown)


mukiwu:
Addressed the review: the watcher now also rescans on Modify(Name) (rename-into-dir) so a new transcript isn't missed if its Create event is dropped/coalesced, while plain Modify(Data) writes still reuse the tracked path. Added a Rust test seeking into the middle of a multibyte char to cover the read_to_string Err -> read_from_start fallback. Todo key stays on item.text (distinct in practice; normalizer has no stable id) (6384391).

PR #31, opened 2026-06-19
feat: redesign Claude progress panel to show activity history and status

mukiwu:
## Summary

Redesigns the right-hand "Claude 進度" panel so it stays useful during normal Claude sessions instead of showing "目前沒有 Claude 活動" almost all the time.

### Root cause

The panel only tracked `runningTools`, which were dropped the instant a tool finished. A typical session (tools that run and complete, no sub-agent, no TodoWrite) left the panel empty, and Opus' long thinking periods kept it empty even longer. A second, independent cause is also fixed: the backend hardcoded `~/.claude/projects/` and ignored `CLAUDE_CONFIG_DIR`.

### Changes

- **Activity history**: `runningTools` becomes a persistent `activities` list. Finished tools stay, marked done/error, capped at 30, so the panel shows recent work even while Claude is thinking. True emptiness now only means the session never ran a tool.
- **Status indicator**: each session header shows a derived status (正在執行 / 思考中 / 等待輸入).
- **CLAUDE_CONFIG_DIR**: the backend honors the env var when locating transcripts, falling back to `~/.claude` (decision extracted into a pure, unit-tested function).
- normalizer, sub-agent detection, and todo logic are intentionally unchanged.

### Out of scope

- workflow / background-worktree agents (they run in directories with no terminal pane; tracked separately)
- parsing background sub-agent completion notifications (background agents stay shown as in-progress)

## Test plan

- [x] Frontend unit tests: 383 passing (51 files), including new activity-history reducer, `deriveStatus`, and store selector tests
- [x] `tsc --noEmit` clean
- [x] Backend `cargo test`: 86 passing, including new `config_base_dir` tests
- [ ] Manual in-app: run interactive `claude` in a pane under the default `~/.claude` config, have it use several tools, and confirm the panel lists the activity history, keeps finished tools, persists through thinking, and shows the correct header status

### DIFF
--- a/src-tauri/src/modules/claude_progress/mod.rs
+++ b/src-tauri/src/modules/claude_progress/mod.rs
@@ -118,9 +118,17 @@ fn mangle_cwd(cwd: &str) -> String {
 
 /// The base config directory Claude Code writes transcripts under: the
 /// `CLAUDE_CONFIG_DIR` override when set and non-empty, otherwise `~/.claude`.
+/// A leading `~` in the override is expanded against `home`; shells usually
+/// expand it first, but a literal `~` can still reach us from a config file.
 fn config_base_dir(home: &Path, env_value: Option<&str>) -> PathBuf {
     match env_value {
-        Some(value) if !value.trim().is_empty() => PathBuf::from(value),
+        Some(value) if !value.trim().is_empty() => {
+            let path = Path::new(value);
+            match path.strip_prefix("~") {
+                Ok(rest) => home.join(rest),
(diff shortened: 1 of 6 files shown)


mukiwu:
Thanks — adopted in 14ded21. `config_base_dir` now expands a leading `~` against the home directory (absolute paths, including shell-expanded ones, are unaffected), covered by a new test `config_base_dir_expands_a_leading_tilde_against_home`. The original "no expansion" choice rested on an unverified assumption about Claude Code's behavior; your point is correct.

PR #40, opened 2026-06-22
feat: track Codex CLI session status and progress

mukiwu:
## Summary

Adds live session tracking for **Codex CLI**, at parity with the existing Claude Code tracking: a status badge (執行中 / 思考中 / 等待批准 / 等待輸入) and the full progress panel (tool activities + plan todos). Built on the insight that the existing pipeline is already agent-agnostic over a `ProgressEvent` stream, so Codex lights up the same reducer, store, and panel once its data is normalized into that shape.

Approach C from the design: a shared core with three thin per-agent seams, leaving Claude behavior unchanged.

## What changed

**Badge layer (hook → OSC, reuses the existing mechanism)**
- The existing `status-hook.sh` is now installed into Codex's config too. New Rust commands `codex_status_hook_install` / `codex_status_hook_uninstall` merge our entries into `~/.codex/hooks.json` and ensure `[features] hooks = true` in `~/.codex/config.toml`.
- The hook merge was parameterized by event list so Claude and Codex share the same pure merge/remove functions with different event maps (Codex has no `Notification` event).
- The crash backstop and foreground detection learned about Codex (`isCodexForeground` / `isTrackedAgentForeground`), so a Codex badge clears correctly when no tracked agent is foreground.

**Progress layer (file watcher → normalizer → existing panel)**
- New `codex_progress` module: a single recursive `notify` watcher over `~/.codex/sessions/`, which locates each watched cwd's newest rollout by reading the `session_meta` line (Codex stores by date, not by cwd), tails appended lines via the shared tail core, and emits `ProgressBatch { cwd, agent: "codex", lines, reset }`.
- New `createCodexNormalizer` maps Codex rollout JSONL (`function_call` / `custom_tool_call` / `*_output` / `update_plan` / `task_complete`) into the same `ProgressEvent` union the Claude normalizer produces.
- `progressStore.pushLines` now selects the normalizer per agent and rebuilds it on an agent switch; `App.tsx` routes each batch by its `agent` tag.
- Workspace cards show a small `Claude` / `Codex` label next to the badge, sourced from the agent feeding the panel.

## Coexistence & safety

- **Never clobbers CodeIsland**: our hook entries are identified by our script path, so a pre-existing `codeisland-bridge --source codex` entry survives both install and uninstall (covered by a test).
- **Preserves the user's config**: `config.toml` is edited with `toml_edit` (comments, keys, and formatting preserved) and written atomically (temp-then-rename). Uninstall leaves the shared `hooks` flag in place.

## Out of scope (deferred by design)

- Codex subagents in the panel (Codex stores child agents as separate threads; structurally different from Claude's inline model).
- Antigravity and Gemini.

## Known follow-up (not a blocker)

- `codex_progress::route_event` rescans the today+yesterday rollout dirs on every filesystem event. The Claude watcher gates rescans on `Create`/`Rename` event kinds; Codex should do the same (or cache the cwd→path map) before this scales to heavy users. Tracked as a follow-up.

## Test plan

**Automated (passing):**
- [x] Rust: `cargo test` — 123 passed (incl. Codex normalizer-adjacent hook/merge/toml/parse/scan/select tests)
- [x] Frontend: `pnpm test` — 512 passed; `pnpm run typecheck` clean
- [x] Per-task reviews + a final whole-branch review (verdict: ready to merge, 0 Critical / 0 Important)

**Manual verification (TODO before merge):**
- [ ] Run the app with tracking on; confirm `~/.codex/hooks.json` still contains CodeIsland's entries **and** ours, and `~/.codex/config.toml` keeps all prior keys with `[features] hooks = true`
- [ ] Launch `codex` in a pane; badge transitions through thinking / active / waiting-approval / idle; a plain shell shows no badge
- [ ] During a Codex turn, the panel shows tool activities (Shell, Edit, Read) and, on `update_plan`, the todo list
- [ ] Switching a directory between Codex and Claude sessions resets the panel cleanly
- [ ] Claude sessions behave exactly as before
- [ ] CPU stays reasonable under a busy Codex session (watch the full-tree-rescan follow-up above)

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1571,7 +1571,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0bb0228f477c0900c880fd78c8759b95c7636dbd7842707f49e132378aa2acdc"
 dependencies = [
  "heck 0.4.1",
- "proc-macro-crate 2.0.2",
+ "proc-macro-crate 2.0.0",
  "proc-macro-error",
  "proc-macro2",
  "quote",
@@ -3209,11 +3209,10 @@ dependencies = [
 
 [[package]]
 name = "proc-macro-crate"
-version = "2.0.2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b00f26d3400549137f92511a46ac1cd8ce37cb5598a96d382381458b992a5d24"
+checksum = "7e8366a6159044a37876a2b9817124296703c586a5c92e2c53751fa06d8d43e8"
 dependencies = [
- "toml_datetime 0.6.3",
  "toml_edit 0.20.2",
 ]
 
@@ -4703,6 +4702,7 @@ dependencies = [
  "tauri-plugin-updater",
  "tauri-plugin-window-state",
  "tokio",
+ "toml_edit 0.22.27",
  "trash",
 ]
 
@@ -4856,7 +4856,7 @@ checksum = "185d8ab0dfbb35cf1399a6344d8484209c088f75f8f68230da55d48d95d43e3d"
 dependencies = [
  "serde",
  "serde_spanned 0.6.9",
- "toml_datetime 0.6.3",
+ "toml_datetime 0.6.11",
  "toml_edit 0.20.2",
 ]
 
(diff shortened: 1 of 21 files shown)


mukiwu:
Thanks for the review — all four suggestions adopted:

- **(high) Watcher rescan optimization** `codex_progress/mod.rs`: `route_event` now takes the notify event and only rescans the sessions tree on Create/rename (or when a watched cwd has no tracked file yet); plain `Modify(Data)` appends tail the already-tracked file directly, matching the Claude watcher. → commit 7065574
- **(medium) Null-safe normalizer** `codexNormalize.ts`: `record?.payload` (guards `JSON.parse("null")` → null, which sits outside the parse try/catch) and `s?.step`/`s?.status` in `parsePlan`. Added tests for a `null` line and a null plan element. → commit 1e4da6d

All green after the changes: Rust `cargo test` and frontend `pnpm test` (514) + `tsc --noEmit` clean.

PR #41, opened 2026-06-22
feat: list every agent session per pane in the workspace sidebar

mukiwu:
## Problem

A tab split across two panes (e.g. Codex CLI in one pane, Claude Code in another) only showed a single session in the Workspace sidebar. Two causes stacked up:

- The card rendered one row per tab, keyed to a single derived cwd (`deriveTabCwd`), so only the active pane's session was shown.
- Progress, agent label, and title were all tracked per-cwd, so two agents running in the **same** directory clobbered each other (every agent switch reset the other's normalizer and overwrote its label).

## Changes

- **Per-(cwd, agent) progress** — `progressStore` keys `sessions`, `sessionEpochs` (and the normalizers) by `progressKey(cwd, agent)`, so Claude and Codex in the same directory stay independent. No backend change needed; the watcher already emits both streams tagged with their agent.
- **Per-pane agent** — each terminal pane's agent is classified from its foreground process (reusing the existing 1.2s poll) and stored per leaf in `sessionStatusStore` next to the live status.
- **Multi-session card** — a tab running 2+ agents shows the tab name as the header and one row per pane, each with its own status badge, agent label, and title. Single-session cards are unchanged.
- **Codex titles** — new `codex_session_title` Tauri command reads the newest Codex rollout's first `user_message` (skipping injected `environment_context`), mirroring `claude_session_title`. `titlesStore` is keyed by `(cwd, agent)` and dispatches to the right source.

## Test plan

- [x] `pnpm test` — 534 passing, including a new WorkspacePanel acceptance test that renders a split tab and asserts both Claude and Codex rows appear
- [x] `pnpm typecheck` clean
- [x] `cargo build` + `cargo test` for `codex_session_title` / `extract_codex_title`
- [x] Local signed `pnpm tauri build` succeeds
- [ ] Manual: split a tab, run `codex` in one pane and `claude` in the other (same dir and different dirs), confirm both sessions list in the sidebar with correct labels/titles

## Known limitations

- Two panes running the **same** agent in the **same** directory share one title/progress slot.
- A multi-session card still shows branch/path for the primary pane only (per-row branch not yet rendered).
- The agent label appears after the foreground poll classifies the pane (up to ~1.2s).

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -11,7 +11,7 @@ use modules::ai::ai_chat;
 use modules::claude_progress::{
     claude_progress_unwatch, claude_progress_watch, claude_session_title, ClaudeProgressState,
 };
-use modules::codex_progress::CodexProgressState;
+use modules::codex_progress::{codex_session_title, CodexProgressState};
 use modules::claude_status_hook::{claude_status_hook_install, claude_status_hook_uninstall};
 use modules::codex_status_hook::{codex_status_hook_install, codex_status_hook_uninstall};
 use modules::notes::{notes_unwatch, notes_watch, NotesWatchState};
@@ -163,6 +163,7 @@ pub fn run() {
             claude_progress_watch,
             claude_progress_unwatch,
             claude_session_title,
+            codex_session_title,
             claude_status_hook_install,
             claude_status_hook_uninstall,
             codex_status_hook_install,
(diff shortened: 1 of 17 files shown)


mukiwu:
Thanks for the review — both points addressed in 501a685:

- **`basename` Windows paths**: now splits on both separators via `path.split(/[/\\]/)`. (Used a corrected regex; the suggested snippet's character class was malformed.)
- **`extract_codex_title` lazy read**: now takes a `BufRead` reader and stops at the first `user_message`, so a long rollout isn't read into memory. `codex_session_title` passes a `BufReader<File>`; tests pass `contents.as_bytes()`.

PR #42, opened 2026-06-22
feat: highlight the current commit in the git graph

mukiwu:
## What

The git graph had no persistent cue for which commit `HEAD` points at — only the small green head-ref chip. This marks the current commit and tidies two lane colours.

- **Current commit highlight (Variation B — node halo)**: the HEAD commit's graph node is filled with `--color-accent` (a colour the branch lanes never use) and given a soft, reduced-motion-aware breathing glow. The commit rows stay calm — the current row only brightens to full foreground — so the list still reads uniformly and the cue doesn't fight row selection.
- **`isCurrentCommit()`**: a pure helper — a commit is current when it carries a `head` ref. Covers both an attached HEAD (`HEAD -> branch`) and a detached HEAD (bare `HEAD`).
- **orange / yellow separation**: the two warm lane colours were only ~12° of hue apart (and the yellow had been deepened toward gold), so adjacent lanes looked alike. orange `#e08a3c` → `#e3712f` (redder), yellow `#d6a82c` → `#e8c139` (brighter gold).

## Test plan

- [x] `pnpm test` — 538 passing, incl. new `isCurrentCommit` tests (attached HEAD, detached HEAD, non-head refs, no refs)
- [x] `pnpm typecheck` clean
- [x] Local signed `pnpm tauri build` succeeds
- [ ] Manual: open the git graph, confirm the HEAD node glows in accent and rows stay calm; check orange vs yellow lanes are distinct on both dark and light themes

## Notes

- The glow honours `prefers-reduced-motion: reduce` (animation disabled).
- Styles were chosen via standalone HTML previews (Variation A/B/C, then HEAD-node colour, then the orange/yellow pair) to avoid repeated app rebuilds.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -275,3 +275,30 @@ button:disabled {
 .note-md .hljs-variable, .note-md .hljs-template-variable,
 .note-md .hljs-property, .note-md .hljs-params { color: var(--color-fg); }
 .note-md .hljs-meta, .note-md .hljs-comment { color: var(--color-fg-subtle); }
+
+/*
+ * Git graph: the current (HEAD) commit's node. Filled with the accent — a colour
+ * the branch lanes never use — with a soft glow that gently breathes, so "you are
+ * here" is unmistakable without disturbing the calm commit rows.
+ */
+.git-head-node {
+  box-shadow:
+    0 0 0 1.5px var(--color-accent),
+    0 0 11px 2px color-mix(in srgb, var(--color-accent) 60%, transparent);
+  animation: git-head-pulse 2.6s ease-in-out infinite;
+}
+@keyframes git-head-pulse {
+  0%, 100% {
+    box-shadow:
+      0 0 0 1.5px var(--color-accent),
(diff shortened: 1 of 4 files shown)


mukiwu:
Thanks for the review.

**Addressed (201a64f):**
- **`bg-accent` for the HEAD node** — switched the current commit's node from an inline `var(--color-accent)` to the `bg-accent` utility; the inline style now only carries the branch hex for non-current nodes.

**Not taking — `isCurrentCommit` optional chaining + undefined tests:**
- `isCurrentCommit(commit: CommitNode)` takes a typed `CommitNode`, and `refs` is a Rust `Vec<GraphRef>` that always serializes as an array (`[]` when empty), never null/undefined. The only caller maps over a non-null `CommitNode[]`, so `commit` is never undefined either.
- Adding `commit?.refs?. … ?? false` would guard states the types forbid, and the `as any` tests would assert inputs that can't occur — they'd pass without protecting against anything real. I'd rather keep the helper's contract honest. Happy to revisit if the graph payload ever becomes untrusted.

PR #43, opened 2026-06-22
fix: restore changelog notes in latest.json after Windows CI

mukiwu:
## Bug

The in-app updater's "更新內容" dialog showed up **empty** for v0.0.7 (just a dash).

## Root cause

The updater renders the `notes` field of `latest.json` (not the GitHub release body). The release flow is: the macOS `release.sh` builds `latest.json` with `notes` embedded from `CHANGELOG-NEXT.md`, then the `v*` tag push triggers `windows-build.yml`, whose `tauri-action` **rewrites** `latest.json` to add the Windows platforms — merging the platform entries correctly but wiping `notes` to an empty string.

Evidence on the v0.0.7 release: `release.sh`'s local manifest had `notes` of 1267 chars (pub_date 16:58); the published manifest had `notes: ""` with pub_date 17:03, exactly during the Windows CI run.

(The live v0.0.7 `latest.json` has already been repaired out-of-band by re-uploading it with `notes` restored.)

## Fix

Add a post-build step in `windows-build.yml` (tag pushes only) that, after `tauri-action`, downloads `latest.json`, re-injects `notes` from this tag's `CHANGELOG-NEXT.md` (still populated at the tagged commit — it's only reset in a later commit), and re-uploads it. The JSON patching is a small `scripts/patchManifestNotes.mjs` helper.

- A blank changelog leaves existing notes untouched (a missing file can't blank a good manifest).
- The ref is passed via an `env:` var and quoted in the `run:` block to avoid shell injection from a crafted tag name.

## Test plan

- [x] `pnpm test` — new `patchManifestNotes` tests (restore notes + keep platforms; blank changelog is a no-op)
- [x] Workflow YAML parses
- [ ] Next release: confirm the published `latest.json` keeps both the merged platforms and the changelog `notes`

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -75,6 +75,24 @@ jobs:
           prerelease: false
           args: "--target x86_64-pc-windows-msvc"
 
+      - name: Restore changelog notes in latest.json
+        # tauri-action above merges the Windows platforms into the release's
+        # latest.json, but it rewrites the file and drops the `notes` the macOS
+        # release embedded from the changelog. The in-app updater renders `notes`,
+        # so put it back from this tag's CHANGELOG-NEXT.md (still populated at the
+        # tagged commit; it's only reset in a later commit).
+        if: startsWith(github.ref, 'refs/tags/')
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


mukiwu:
Addressed in 63d40a6: the CLI now guards the changelog read with `existsSync`, so a missing CHANGELOG-NEXT.md is treated as a blank one (no-op, existing notes left intact) instead of throwing ENOENT and failing the release step.

PR #56, opened 2026-06-26
feat(terminal): add in-terminal search (find next/previous)

mukiwu:
## Summary

The terminal had no way to search its scrollback — `@xterm/addon-search` was already a dependency but was never wired up. This adds a search bar.

- **Open**: `Cmd+F` (macOS) / `Ctrl+Shift+F` (other platforms). `Ctrl+F` is intentionally avoided off-mac because it is readline's forward-char binding.
- **Navigate**: `Enter` next, `Shift+Enter` previous; the up/down buttons do the same.
- **Close**: `Escape` or the close button — also clears the match highlight.
- Scoped to the focused pane; opening only affects the terminal that has focus.

## Implementation

| Change | File |
|---|---|
| Load `SearchAddon`, expose it on `TerminalHandle` | `terminal/lib/createTerminal.ts` |
| Search bar component (input + prev/next/close, autofocus, i18n) | `terminal/SearchBar.tsx` |
| `Cmd/Ctrl+Shift+F` interception + overlay render | `terminal/TerminalView.tsx` |
| Shortcut entry | `settings/ShortcutsSettingsSection.tsx` |
| Strings | `i18n/locales/{en,zh-Hant}/{common,settings}.json` |

`SearchBar` depends on a minimal `TerminalSearchController` interface (`findNext` / `findPrevious` / `clearDecorations`) that the real `SearchAddon` satisfies structurally, which keeps the component unit-testable with a fake.

Match highlighting currently uses xterm's native selection (the found text is selected and scrolled into view). Persistent multi-match decoration coloring is a possible follow-up.

## Test plan

- [x] `createTerminal.test.ts` — the handle exposes a working search addon (finds present text, misses absent text)
- [x] `SearchBar.test.tsx` — Enter→next, Shift+Enter→previous, Escape→close, unmount→clear highlight, next/previous buttons
- [x] Full suite green (693 tests), `tsc --noEmit` clean
- [ ] Manual smoke test of the `Cmd+F` interaction in the running app

### DIFF
--- a/src/i18n/locales/en/common.json
+++ b/src/i18n/locales/en/common.json
@@ -3,6 +3,12 @@
   "tagline": "AI-powered terminal workspace",
   "terminalConnecting": "Starting shell…",
   "openLinkHint": "{{mods}}-click to open",
+  "terminalSearch": {
+    "placeholder": "Find in terminal",
+    "next": "Find next",
+    "previous": "Find previous",
+    "close": "Close search"
+  },
   "ssh": {
     "disconnected": "Disconnected — click to reconnect",
     "reconnect": "Reconnect"
--- a/src/i18n/locales/en/settings.json
+++ b/src/i18n/locales/en/settings.json
@@ -134,6 +134,7 @@
     "splitRight": "Split left / right",
     "splitDown": "Split top / bottom",
     "cyclePane": "Cycle split panes",
+    "searchTerminal": "Search terminal output",
     "lineStart": "Jump to line start",
     "lineEnd": "Jump to line end",
     "wordBack": "Jump word left",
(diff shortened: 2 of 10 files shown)


mukiwu:
Thanks for the review — all three findings addressed:

- **High (paste into search bar):** the document-level paste interceptors now ignore events whose target is an `<input>`, so pasting into the search field works instead of being redirected to the shell.
- **Medium (re-press shortcut):** pressing the search shortcut while the bar is open now refocuses and selects the input rather than being a no-op.
- **Medium (clear-on-empty):** emptying the query clears the match highlight immediately. Test added.

PR #57, opened 2026-06-26
feat(terminal): hover action cards for IPs, host:port and archives

mukiwu:
## Summary

Hovering an entity in terminal output now pops a small card of quick commands you can run with one click — inspired by nyaterm's ActionLinks.

| Entity | Actions |
|---|---|
| IPv4 (`10.0.0.1`) | `ping` · `traceroute` · `ssh` · `curl http://…` |
| host:port (`example.com:8080`, `192.168.1.1:8080`) | `curl http://…` · `curl https://…` · `nc host port` · `telnet host port` |
| archive (`backup.tar.gz`, `data.zip`, `blob.7z`, …) | extension-correct **extract** + **list** (`unzip` / `tar -xzf` / `7z x` / …) |

**Clicking an action runs it directly** in the focused terminal. As a safety net, a command that looks destructive (`rm -rf`, `dd … of=`, `mkfs`, pipe-to-shell, fork bomb, raw-disk overwrite) shows a **red confirmation** before running. A settings toggle (`Hover action cards`, on by default) turns the whole thing off.

## Implementation

| Piece | File | Tested |
|---|---|---|
| Matchers (IP / host:port / archive) with overlap resolution | `terminal/lib/actionLinks.ts` | ✅ pure unit tests |
| Per-entity command builders | same | ✅ |
| Destructive-command detection | same | ✅ |
| Hover card UI + red confirm flow | `terminal/ActionCard.tsx` | ✅ RTL |
| Link-provider wiring + hover show/hide + run | `terminal/TerminalView.tsx` | glue |
| Setting + toggle | `stores/settingsStore.ts`, `settings/TerminalSettingsSection.tsx` | — |
| Strings | `i18n/locales/{en,zh-Hant}/{common,settings}.json` | — |

Notes:
- host:port deliberately requires a dotted host or `localhost`, so `error:42`-style key:number pairs are not matched.
- An archive filename also matches as a file path; the overlapping file link is dropped so the action card wins (opening a binary archive as a file isn't useful).
- The card escapes the pane via `position: fixed`; a short hide delay + cancel-on-card-enter lets the pointer travel from the link into the card.

## Test plan

- [x] `actionLinks.test.ts` — 11 cases: each matcher, IP-with-port precedence, `error:42` rejection, per-extension extract/list, danger detection, safe-command pass
- [x] `ActionCard.test.tsx` — safe action runs immediately; dangerous action requires confirm; cancel does not run
- [x] Full suite green (701 tests), `tsc --noEmit` clean, `vite build` clean
- [ ] Manual smoke test in the running app: hover an IP / archive in real output, run an action, confirm the destructive-command warning

### DIFF
--- a/src/i18n/locales/en/common.json
+++ b/src/i18n/locales/en/common.json
@@ -3,6 +3,21 @@
   "tagline": "AI-powered terminal workspace",
   "terminalConnecting": "Starting shell…",
   "openLinkHint": "{{mods}}-click to open",
+  "actionLinks": {
+    "hoverHint": "Hover for actions",
+    "ping": "Ping",
+    "traceroute": "Traceroute",
+    "ssh": "SSH",
+    "curl": "curl (HTTP)",
+    "curlHttps": "curl (HTTPS)",
+    "nc": "Netcat",
+    "telnet": "Telnet",
+    "extract": "Extract",
+    "list": "List contents",
+    "dangerWarning": "This command looks destructive.",
+    "runAnyway": "Run anyway",
+    "cancel": "Cancel"
+  },
   "ssh": {
     "disconnected": "Disconnected — click to reconnect",
     "reconnect": "Reconnect"
--- a/src/i18n/locales/en/common.json
+++ b/src/i18n/locales/en/common.json
@@ -5,6 +5,7 @@
   "openLinkHint": "{{mods}}-click to open",
   "actionLinks": {
     "hoverHint": "Hover for actions",
+    "preview": "Open in preview",
     "ping": "Ping",
     "traceroute": "Traceroute",
     "ssh": "SSH",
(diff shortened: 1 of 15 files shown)


mukiwu:
Thanks for the review — all five findings addressed:

- **High (shell injection):** archive filenames are now single-quoted before building extract/list commands. (The matcher only captures `[A-Za-z0-9_.-]`, so it was not currently reachable, but quoting makes it robust against future matcher changes.)
- **Medium (overlap resolution):** replaced the IP-specific filter with a generic longest-match-wins pass, so `1.2.3.4.zip` resolves to the archive. Test added.
- **Medium (confirm dialog):** the destructive-command confirmation now shows the full command (wrap + scroll) instead of truncating it.
- **Medium (rm long options):** danger detection now also matches `--recursive` / `--force`. Test added.
- **Medium (timer leak):** the pending hover-card hide timer is cleared on unmount.

PR #61, opened 2026-06-26
feat(terminal): large-output protection (batched writer + overload notice)

mukiwu:
## Summary

Terminal output was written to xterm synchronously on every PTY/SSH data event (`onData → term.write`), so a flood (cat a huge file, runaway logs, `yes`) could block the UI thread and freeze the window. This routes live output through a frame-scheduled writer with overload protection, inspired by nyaterm's pipeline.

Built in three slices:

1. **Batched output writer** (`terminal/lib/outputWriter.ts`) — queues chunks and flushes on `requestAnimationFrame` in bounded per-frame chunks (32 KB), dropping the oldest bytes once the backlog passes a cap (1 MB). `requestAnimationFrame`/`cancel` are injectable, so the queue/throttle/drop logic is fully unit-tested.
2. **Wiring** — both PTY and SSH `onData` paths push into the writer; it's disposed on pane teardown. Restored-history and error notices still write directly, so their ordering is unchanged.
3. **Overload notice** — the writer reports a running dropped-bytes total via `onDrop`; a transient notice (throttled to avoid render churn during a flood, auto-hiding once output settles) tells the user how much was skipped.

Constants follow nyaterm's validated values: 32 KB per frame, 1 MB backlog cap.

## Test plan

- [x] `outputWriter.test.ts` — buffers + flushes on frame, per-frame byte budget + reschedule, drop-oldest past the cap, `onDrop` running total, no writes after dispose
- [x] Full suite green (738), `tsc --noEmit` clean, `vite build` clean
- [x] Manual: normal interaction (typing, vim/htop) feels unchanged; a flood (`yes | head -n 5000000`, big `cat`) keeps the UI responsive instead of freezing
- [ ] Manual: the overload notice appears during a flood showing the skipped amount and auto-hides once output settles

### DIFF
--- a/src/i18n/locales/en/common.json
+++ b/src/i18n/locales/en/common.json
@@ -2,6 +2,7 @@
   "appName": "TempoTerm",
   "tagline": "AI-powered terminal workspace",
   "terminalConnecting": "Starting shell…",
+  "outputThrottled": "Output throttled, skipped {{size}} to keep the UI responsive",
   "openLinkHint": "{{mods}}-click to open",
   "actionLinks": {
     "hoverHint": "Hover for actions",
--- a/src/i18n/locales/zh-Hant/common.json
+++ b/src/i18n/locales/zh-Hant/common.json
@@ -2,6 +2,7 @@
   "appName": "TempoTerm",
   "tagline": "AI 驅動的終端機工作區",
   "terminalConnecting": "啟動中…",
+  "outputThrottled": "輸出太快，已略過 {{size}} 以維持介面流暢",
   "openLinkHint": "{{mods}} + 點擊開啟",
   "actionLinks": {
     "hoverHint": "可執行的動作",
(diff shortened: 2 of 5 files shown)


mukiwu:
Thanks for the review.

**Applied (high):** the overload notice now tracks a baseline of previously dropped bytes and displays `total - baseline`, so each overload event shows only its own skipped amount rather than the writer's lifetime cumulative total. Fixed in the latest commit.

**Not applied (the `noteDroppedOutputRef` indirection):** `noteDroppedOutput` closes over only stable values — `useRef` handles (`skippedTotalRef`, `skippedBaselineRef`, the two timer refs) and the stable `setOutputSkipped` setter. It reads no per-render state or props, so the first-render closure captured by the mount-only effect behaves identically to any later one; there's no stale-closure risk to guard against. Wrapping it in a mutable ref would add indirection without changing behaviour, so I kept the direct call.

Full suite green (738), `tsc --noEmit` clean.

PR #62, opened 2026-06-26
feat(terminal): per-line timestamp gutter (toggleable)

mukiwu:
## Summary

A toggleable left gutter showing each output line's write time (`HH:MM:SS`, absolute, 24h). Off by default; built on the batched output pipeline from #61. Brainstormed and spec'd first (`docs/superpowers/specs/` — local, gitignored).

Three components:

1. **`lineTimestamps`** (`terminal/lib/lineTimestamps.ts`) — records buffer line → write time; first write wins; oldest entries pruned past a cap. Pure, unit-tested.
2. **`gutterLines` + `TerminalGutter`** (`terminal/lib/gutterLines.ts`, `terminal/TerminalGutter.tsx`) — `computeVisibleStamps` resolves each visible row to its timestamp (wrapped continuation rows resolve to their logical line); the component renders the column aligned to xterm rows, recomputing on a frame via `onRender/onScroll/onResize/onWriteParsed`. Pure helpers unit-tested; pixel alignment verified in-app.
3. **Setting + wiring** — `showTimestamps` (default off) + toggle in Terminal settings; stamping armed at session connect (so restored history stays unstamped); terminal insets by a fixed gutter width and re-fits on toggle.

## Behaviour

- Default off; user opts in
- Every visible line shows its time (same-second lines repeat, so scrolling never loses the reference)
- Live output only — restored-history lines stay blank
- No line numbers, no relative time (out of scope)

## Test plan

- [x] `lineTimestamps.test.ts` — range stamp, first-write-wins, prune
- [x] `gutterLines.test.ts` — `formatClock` HH:MM:SS, visible-row resolution, wrapped continuation
- [x] `settingsStore.test.ts` — `showTimestamps` default off + toggle
- [x] Full suite green (745), `tsc --noEmit` clean, `vite build` clean
- [ ] **Manual (the unverified part): toggle on → times line up with rows across scroll / font-size / theme; terminal re-fits cleanly on toggle; restored history blank; toggle off restores full width**

### DIFF
--- a/src/i18n/locales/en/settings.json
+++ b/src/i18n/locales/en/settings.json
@@ -89,6 +89,8 @@
     "suggestionsHint": "Show a dim completion from your shell history as you type; press → to accept",
     "actionLinks": "Hover action cards",
     "actionLinksHint": "Hover an IP, host:port or archive in the output to get quick commands (ping, curl, extract…)",
+    "timestamps": "Show line timestamps",
+    "timestampsHint": "Show each output line's write time (HH:MM:SS) in a left gutter; the terminal narrows to make room while it's on",
     "clearHistory": "Clear saved history",
     "historyCleared": "Cleared"
   },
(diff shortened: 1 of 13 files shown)


mukiwu:
Dropping this — per-line timestamps belong in the shell prompt (powerlevel10k time segment / zsh RPROMPT '%D{%H:%M:%S}'), which is native, cross-terminal and more useful (it timestamps when a command was run). Aligning a per-row gutter to xterm proved finicky for little gain.

PR #71, opened 2026-06-27
feat(editor): reload open files when they change on disk

### LINKED ISSUE #70: Reload an open file when it changes on disk (e.g. edited by an AI)
## Problem

When a file is open in the editor and something else changes it on disk — an AI agent (Claude / Codex) editing it, a `git pull`, another tool — the editor keeps showing the stale content. Today the only way to see the new content is to close the tab and reopen it.

## Goal

Let the user pick up on-disk changes to an open file without closing/reopening it.

## Option A — Manual "Refresh" button

Add a refresh button to the editor (toolbar / tab) that re-reads the file from disk and replaces the buffer.

- Simple and predictable; the user decides when to reload.
- Reuses `editorStore.load(path, content)` (resets content + baseline → clean) and the `fsReadFile` bridge.
- Downside: the user has to notice the file changed and click.

## Option B — Auto-reload via a file watcher

Watch the open file(s) and reload automatically when the on-disk content changes.

- No manual step; the editor always reflects disk — ideal for the "AI just edited this file" workflow.
- Infra already exists: the notes module uses the `notify` crate (`RecommendedWatcher`, recursive) and pushes change events to the frontend — see `src-tauri/src/modules/notes/mod.rs`. The same pattern can watch open editor files.
- Main risk: **unsaved local edits**. If the buffer is dirty (`editorStore.isDirty`: `content !== baseline`) and the file also changed on disk, auto-overwriting would silently lose the user's edits.

## Recommendation: do both, layered

1. A watcher detects the external change (Option B infra).
2. If the buffer is **clean** → reload silently (optionally with a subtle "reloaded" hint).
3. If the buffer is **dirty** → do NOT overwrite. Show a non-destructive banner ("This file changed on disk") with a **Reload** button (Option A), letting the user choose — mirroring how VS Code handles external changes.

This gives the AI-edit workflow the "just refresh" behavior, while never clobbering unsaved work. A minimal first step can ship Option A (the button) alone, then layer the watcher on top.

## Pointers
- Editor buffers + dirty model: `src/modules/editor/store/editorStore.ts` (`load`, `setContent`, `isDirty`)
- Existing fs-watch pattern to reuse: `src-tauri/src/modules/notes/mod.rs` (`notify` v6 recursive watcher → frontend reload)
- File read bridge: `fsReadFile` / `fs_read_file`

mukiwu:
Closes #70. Pick up on-disk changes to an open file without closing/reopening the tab — both manually and automatically.

## What's in it

- **Manual refresh button** (toolbar): reloads the file from disk. A clean buffer reloads immediately; unsaved edits confirm first (`manualReloadAction`).
- **File watcher** (backend, `notify`): `editor_watch_set(paths)` subscribes to each open file's parent directory (non-recursive, so atomic temp+rename saves are seen) and emits `editor-file-changed`. Reuses the notes watcher's event filter (skip access / metadata to avoid reload loops).
- **Auto-reload + conflict banner** (frontend): a clean buffer reloads silently; unsaved edits raise a non-destructive banner to pick the disk version or keep yours (`externalChangeAction`). The echo of our own save is ignored via a 2s self-write window. `installEditorWatchSync` keeps the watched set in step with the open editor tabs.

Layered exactly as discussed on the issue: watcher detects → clean auto-reloads → dirty is never clobbered.

## Test plan

- [x] New pure-logic tests: `manualReloadAction`, `externalChangeAction` (reload.test.ts), `is_interesting_change` (Rust)
- [x] Full frontend suite — 791 passed
- [x] Full Rust lib tests — 174 passed
- [x] `tsc --noEmit` typecheck passes

## Notes

- The CodeMirror component itself isn't unit-tested (project convention is to test the lib logic); the decision functions are fully covered.
- Cursor/scroll preservation on auto-reload is best-effort for now and can be refined if needed.

### DIFF
--- a/CHANGELOG-NEXT.md
+++ b/CHANGELOG-NEXT.md
@@ -1,11 +1,13 @@
 ## 正體中文
 
 ### feat
+- 編輯器檔案重新整理：開啟的檔案被外部或 AI 改動後不用關掉重開，工具列多了一個重新整理鈕可隨時載入最新內容，若有未存修改會先確認；編輯器也會自動偵測磁碟變更，沒有未存修改時自動重載，有的話跳出提示讓你選要用磁碟版本還是保留自己的
 
 ### fix
 
 ## English
 
 ### feat
+- Editor file reload: when an open file changes on disk (e.g. an AI agent edits it), pick up the new content without closing and reopening the tab. A toolbar refresh button reloads on demand (confirming first if you have unsaved edits), and the editor also watches the file: a clean buffer reloads automatically, while unsaved edits raise a banner to choose between the disk version and your own
 
 ### fix
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -46,6 +46,7 @@ use modules::terminal_history::{
     terminal_history_prune, terminal_history_save,
 };
 use modules::sysmon::{system_stats, SysinfoState};
+use modules::editor_watch::{editor_watch_set, EditorWatchState};
 
 #[derive(serde::Serialize)]
 #[serde(rename_all = "camelCase")]
@@ -90,6 +91,7 @@ pub fn run() {
         .manage(CodexProgressState::new())
         .manage(NotesWatchState::new())
         .manage(SysinfoState::new())
+        .manage(EditorWatchState::new())
         .setup(|app| {
             // window-state restores the last size/position, but it can persist a
             // corrupt tiny / off-screen value (observed 360x240 at a negative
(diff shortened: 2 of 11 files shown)


mukiwu:
Addressed in e7b5ea8:
- **high** (self-write race): marker is now set before the async write, with a reset on failure.
- **high** (watcher churn): the backend is only called when the set of open files actually changes.
- **medium** (resilient watch): individual `watch()` errors are ignored so other files keep being watched.

The remaining **medium** (cross-platform path canonicalization: macOS symlinks / Windows case) is tracked in #72 — it needs a canonical↔frontend path map + UNC strip + cross-platform testing, out of scope for this PR.

PR #74, opened 2026-06-27
fix(terminal): open file paths hard-wrapped across two lines

### LINKED ISSUE #73: Open file links that programs hard-wrap across two lines
## Problem

File path links in the terminal don't open when a program **hard-wraps** a long path across two display lines, e.g. Claude Code's `Write(...)` output where `scratchpad` is split into `scr` / `atchpad` with the continuation indented. Alt/Cmd+click does nothing.

## Root cause

It's a **hard wrap** (the program emits `\n` + indentation), so the two pieces are separate *logical* lines in the buffer. `gatherLogicalLine` only stitches xterm **soft-wrap** (`isWrapped`) continuations, so it can't rejoin a hard wrap, and `findFilePaths` sees only half a path per line. Soft-wrapped paths already work; this is specific to hard wraps. Not a cwd bug.

## Approach (checked against Warp via DeepWiki)

Warp detects paths by regex + **filesystem validation as the gatekeeper**: only show a link if the path actually exists (with a cwd file cache). Validation-first is the key insight: a broad/over-eager join is safe because a wrong join won't resolve on disk, so it never becomes a link.

So: rejoin hard-wrapped candidates, then validate existence before making them links.

## Done

- `wrappedPathCandidates(firstLine, nextLine)` pure helper + tests, rejoins a hard-wrapped absolute path; caller must validate.
- Branch `fix/terminal-wrapped-file-links`, commit `142b3b2`. fileLinks suite 20 passed.

## Remaining (integration layer)

1. In the `TerminalView` link provider (`provideLinks`), besides the current single-logical-line detection, also look at the next logical line and feed `wrappedPathCandidates`.
2. **Validate at detection time**: only emit a link for a candidate that exists on disk. To avoid one IPC per hover, cache the cwd file list (Warp-style) and match in memory.
3. **xterm cross-line link range (main risk)**: the link spans two logical lines (start on line N's fragment, end on line N+1's continuation). `provideLinks` is called per row, so the link must be returned on both rows with an aligned range. Needs real in-app click testing, not just unit tests.

## Out of scope

Bare filenames in prose (no path, and the file lives outside the terminal cwd). No path info to resolve and not in cwd, so even Warp's cwd cache wouldn't find it.

## Pointers

- `src/modules/terminal/lib/fileLinks.ts` — `findFilePaths`, `resolveFilePath`, `wrappedPathCandidates` (new)
- `src/modules/terminal/lib/cellPositions.ts` — `gatherLogicalLine` (soft-wrap stitching), `buildCellPositions`
- `src/modules/terminal/TerminalView.tsx` — link provider (`registerLinkProvider` / `provideLinks`), `openFromTerminal` (current click-time validation)

## Estimate

Slice 1 (detection-time validation + cwd cache) M/medium; Slice 2 (cross-line reassembly + range) L/wide. Confidence low (thin calibration, optimistic HPP). Report at `estimates/tempo-term-filelink-wrap-估時報告.md`.

mukiwu:
Closes #73.

## Problem

When a program (e.g. an AI coding agent) hard-wraps a long absolute path across two lines — the first ends mid-path, the next continues it after indentation — the path was not clickable. It's a hard wrap (`\n` + indent), so the two halves are separate logical lines and `gatherLogicalLine` (soft-wrap only) can't rejoin them; `findFilePaths` saw half a path per line.

## Approach

Checked against Warp (via DeepWiki): it detects paths by regex + filesystem validation as the gatekeeper. Applied here:

- `wrappedPathCandidates(firstLine, nextLine)` (pure, tested) rejoins the two halves of a hard-wrapped absolute path.
- The link provider feeds the previous/next logical line into it and, for each half, adds a **single-line** link pointing at the rejoined path.
- **No cross-line range** (an xterm minefield): each half is its own link. `openFromTerminal` validates the rejoined path exists on click, so a broad/wrong join simply won't open.

## Scope

In: the hard-wrapped absolute path case (image 1). Out: bare filenames in prose with no path and living outside the cwd (image 2) — no info to resolve.

## Test plan

- [x] `wrappedPathCandidates` unit tests (rejoin, reject non-path lines)
- [x] fileLinks suite + full frontend suite — 794 passed
- [x] `tsc --noEmit` passes
- [ ] Manual: the xterm cross-line wiring (clicking either half) needs an in-app click test — flagging for review since unit tests can't cover the link provider's xterm integration.

### DIFF
--- a/CHANGELOG-NEXT.md
+++ b/CHANGELOG-NEXT.md
@@ -4,6 +4,7 @@
 - 編輯器檔案重新整理：開啟的檔案被外部或 AI 改動後不用關掉重開，工具列多了一個重新整理鈕可隨時載入最新內容，若有未存修改會先確認；編輯器也會自動偵測磁碟變更，沒有未存修改時自動重載，有的話跳出提示讓你選要用磁碟版本還是保留自己的
 
 ### fix
+- 終端機裡被程式（例如 AI agent）折成兩行的長檔案路徑，現在點上下任一段都打得開，會自動把被斷開的路徑接回來再開
 
 ## English
 
@@ -11,3 +12,4 @@
 - Editor file reload: when an open file changes on disk (e.g. an AI agent edits it), pick up the new content without closing and reopening the tab. A toolbar refresh button reloads on demand (confirming first if you have unsaved edits), and the editor also watches the file: a clean buffer reloads automatically, while unsaved edits raise a banner to choose between the disk version and your own
 
 ### fix
+- File paths that a program (e.g. an AI agent) hard-wraps across two lines in the terminal are now clickable: clicking either half opens the rejoined path
--- a/src/modules/terminal/TerminalView.tsx
+++ b/src/modules/terminal/TerminalView.tsx
@@ -66,7 +66,7 @@ import {
   shouldAttachImage,
   shellQuotePath,
 } from "./lib/terminalClipboard";
-import { buildFileLink, findFilePaths, resolveFilePath } from "./lib/fileLinks";
+import { buildFileLink, findFilePaths, resolveFilePath, wrappedPathCandidates } from "./lib/fileLinks";
 import { actionsFor, findActionLinks, type TerminalAction } from "./lib/actionLinks";
 import { ActionCard } from "./ActionCard";
 import { buildCellPositions, gatherLogicalLine } from "./lib/cellPositions";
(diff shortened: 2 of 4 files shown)


mukiwu:
Addressed in 5df10a3:
- **high** (relative paths collapsed to bogus absolute): `TRAILING_PATH_RE` now matches the whole path-like suffix (>= 1 slash), so relative paths stay relative.
- **high** (range mismatch): the link provider positions the range with the same regex.
- **medium**: added a relative-path rejoin test.

Full frontend suite 795 passed, typecheck clean.

PR #75, opened 2026-06-27
fix(terminal): enable clipboard paste on Windows

craig7351:
## Problem

Terminal paste did nothing on Windows. Three issues compounded:

1. The native `paste` event was suppressed on **all** platforms (`onPasteCapture` always called `preventDefault`), so xterm's built-in paste never fired.
2. The custom `Ctrl/Cmd+V` handler was **macOS-only** (`isTerminalPasteShortcut` returned `null` off mac), so Windows never ran the smart paste path either.
3. The backend clipboard read returned an empty string on non-macOS.

Net result: native paste was blocked, the custom path was mac-gated, and the backend returned nothing — Windows had no working paste at all.

## Changes

- **Route Windows `Ctrl+V`** through the existing smart paste flow (clipboard text wins, else a copied file's path) via the new `IS_WINDOWS` flag. Linux intentionally keeps xterm's built-in paste (no native clipboard backend yet).
- **Read the Windows clipboard** in the backend with `clipboard-win` (synchronous, no subprocess on the hot path): `clipboard_text()` (`CF_UNICODETEXT`) and `clipboard_paths()` (`CF_HDROP` copied-file list).
- **Quote Windows paths with double quotes** (accepted by cmd, PowerShell and git-bash) instead of POSIX single quotes.
- **Switch russh to its `ring` crypto backend.** The default `aws-lc-rs` pulls `aws-lc-sys`, whose Windows build requires NASM; `ring` builds on Windows MSVC with no extra toolchain. This unblocks `cargo build`/`tauri dev` on a stock Windows setup.

## Not included (technical limitation)

Image auto-attach to an image-aware CLI stays macOS/Linux only: `portable-pty` exposes no foreground process on Windows, so the app cannot detect that `claude`/`codex` is running there.

## Test plan

- `pnpm typecheck` — clean
- `pnpm test` — 688/688 pass (incl. new Windows path-quoting test)
- `cargo check` (Windows) — clean, 0 warnings; `aws-lc-sys`/NASM no longer required
- `cargo test --lib clipboard` — 6/6 pass
- Manual (Windows): `Ctrl+V` pastes clipboard text; copying a file in Explorer then `Ctrl+V` pastes its (quoted) path


craig7351:
Pushed an additional commit to this branch: **fast right-click paste menu on Windows** (`8cff0b0`).

Right-click → Paste previously went through the WebView2 native context menu, whose paste command is slow (~5s), while Ctrl+V was instant. This adds a custom Copy/Paste context menu on the terminal (Windows only) that routes paste through the same fast handler as Ctrl+V. macOS and Linux are untouched (they keep their native menus and have no slow-paste issue).

- `pnpm typecheck` ✅ · `pnpm test src/modules/terminal` → 90 passed ✅ · verified with `tauri dev`.

mukiwu:
Rebuilt on top of current `master` (resolved the merge conflict: `master` now carries `IS_WINDOWS` from #67, plus the action-card and search-bar additions to `TerminalView`) and addressed both review comments in 806eefa:

- **High** — the suppressed native `paste` event is now routed through `handleTerminalPaste("cmd")`, so right-click and Edit-menu paste work on macOS and Windows too, not just the keyboard shortcut. Keyboard Ctrl/Cmd+V still `preventDefault`s on keydown, so the native paste event never fires for it and there is no double paste.
- **Medium** — dropped the comma from the bare (unquoted) Windows path character set, so a path containing a comma is double-quoted. PowerShell parses a bare comma as its array operator. Added a regression test.

Verified on macOS: `pnpm typecheck` clean, `pnpm test` 796/796, `cargo build` clean with the `ring` backend, `cargo test --lib clipboard` 7/7.

PR #76, opened 2026-06-27
fix(claude-hook): use forward-slash paths so the status hook runs on Windows

craig7351:
## Problem

The Claude Code status hook never fired on Windows. Claude Code runs `command` hooks through bash, which treats `\` as an escape character. A backslash script path like `C:\Users\...\status-hook.sh` collapses to `C:Users...status-hook.sh`, so the hook command can't be found.

## Changes

- Add a `normalize()` helper that rewrites `\` to `/` on Windows (a no-op on Unix), and store/compare the hook command in that single forward-slash form. Git Bash accepts forward slashes, so the command bash runs is valid on Windows.
- On reinstall, strip stale backslash entries left by older installs so they don't stack alongside the corrected one.

## Test plan

- `cargo test --lib claude_status_hook` — 8/8 pass, including the new `#[cfg(windows)]` `windows_install_dedups_across_slash_styles` test that asserts a stale backslash entry is replaced by exactly one forward-slash entry per event.

### DIFF
--- a/src-tauri/src/modules/claude_status_hook/mod.rs
+++ b/src-tauri/src/modules/claude_status_hook/mod.rs
@@ -40,16 +40,11 @@ fn our_command(script_path: &str, state: &str) -> String {
 /// `command` hooks through bash, which treats `\` as an escape and mangles the
 /// path (`C:\Users\...` collapses to `C:Users...`, so the hook can't be found).
 /// Git Bash accepts forward slashes, so we store and match on a single
-/// forward-slash form. A no-op on Unix, where paths already use forward slashes.
+/// forward-slash form. Applied unconditionally: Unix paths never contain a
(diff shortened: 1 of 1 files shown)


mukiwu:
Addressed all three review comments in 146068d:

- `normalize()` is now unconditional (dropped the `#[cfg]` branches). Unix paths never contain a backslash, so it stays a no-op there, and the dedup logic can now be exercised on any CI runner.
- Removed the redundant `normalize(script_path)` inside `remove_hook_settings`; all callers already pass a normalized path, now documented as a precondition.
- Un-gated `windows_install_dedups_across_slash_styles` so it runs on every platform, not just Windows.

`cargo test --lib claude_status_hook`: 8/8 pass on macOS, including the dedup test that previously only ran on Windows.

PR #91, opened 2026-06-30
fix(tabs): unify sidebar context menu with main TabBar

### LINKED ISSUE #86: 終端機右鍵選單統一
正常終端右鍵會這樣
Image

側欄終端會長這樣
Image

mukiwu:
## Summary

Fixes #86. The sidebar workspace TabCard had no `onContextMenu` handler, so right-clicking fell through to the WebView's native menu (Back / Reload / Print). The main TabBar already showed a custom Rename Tab / Close Tab menu via `ContextMenu`. Both surfaces now share the same menu and the same dirty-tab close confirmation.

- Extract `tabContextMenuItems` so both surfaces stay in lockstep.
- Add `onContextMenu`, inline rename, and the shared menu to sidebar `TabCard`.
- Extract `useTabCloseRequest` hook that owns the dirty check + confirm dialog so closing a dirty tab from the sidebar no longer silently drops unsaved work. Dialog is portaled so consumers can drop the returned node inside any wrapper (including `<button>`).

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 850/850 (added 1 lock-in test for sidebar dirty-close path)
- [ ] Main TabBar right-click → Rename Tab / Close Tab menu, same as before
- [ ] Sidebar TabCard right-click → identical menu (no native WebView menu)
- [ ] Sidebar Rename → inline input, Enter commits, Esc cancels
- [ ] Sidebar Close on a clean tab → tab closes immediately
- [ ] Sidebar Close on a dirty editor tab → Unsaved changes dialog appears, tab does not close until confirmed

### DIFF
--- a/src/components/TabBar.tsx
+++ b/src/components/TabBar.tsx
@@ -7,7 +7,6 @@ import {
   Globe,
   LayoutGrid,
   PanelLeft,
-  Pencil,
   Plus,
   ScrollText,
   SquareTerminal,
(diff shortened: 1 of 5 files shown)


mukiwu:
Thanks for the read. Acknowledging the summary — no inline findings, so no follow-up commit needed on this branch.

PR #105, opened 2026-07-01
fix(pr): stop gh spawns flashing a console and stalling the UI on Windows

yw-chan:
## Problem
On Windows, a black console window flashed repeatedly and the UI could feel stuck. Two Windows-specific costs were behind it:

- The workspace cards poll for each branch's PR on a timer. Every `gh` spawn flashed a fresh console window, and each spawn was a blocking call with no timeout.
- Every open terminal ran a 1.2s cwd-tracking poll that, on Windows, only fired IPC that always came back empty.

## Root cause
- A release build has no console of its own (`windows_subsystem = "windows"`), so any child process spawned without `CREATE_NO_WINDOW` gets a fresh console that flashes. `git::run_git` already set the flag, but the `gh` spawns (`gh_available`, `pr_via_gh`) did not. This only started showing after #89 made `gh` actually resolve on Windows.
- `Command::output()` has no timeout, so a hung `gh` (network stall, an interactive auth prompt) would block the calling thread indefinitely; on a timer these could pile up.
- `pty_cwd` / `pty_foreground_command` have no Windows backend — `read_process_cwd` returns `None` there (no `/proc`, no `lsof`). The frontend still called them every 1.2s per terminal.

## Changes
- **fix(pr):** add a `gh_command` helper that sets `CREATE_NO_WINDOW` on Windows and route both `gh_available` and `pr_via_gh` through it.
- **fix(pr):** add a `run_gh` helper (via the `wait-timeout` crate) that kills a `gh` invocation after 5s, so a hung `gh` can no longer block a caller thread.
- **perf(terminal):** skip the cwd-tracking poll entirely on Windows, since its backend is a no-op there. The guard should be dropped once a Windows cwd implementation lands.

## Test plan
- [x] `npx pnpm exec tsc --noEmit` — passes
- [x] `npx pnpm exec vitest run` — 872 passed (127 files)
- [x] `cargo test --lib pr::` — 12 passed
- [x] `cargo check` — clean
- [x] Manual: built a release bundle, opened workspace cards bound to git repos and toggled window focus to trigger PR polling — no console windows flash; terminals open without the earlier stall.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -6023,6 +6023,7 @@ dependencies = [
  "tokio",
  "toml_edit 0.22.27",
  "trash",
+ "wait-timeout",
 ]
 
 [[package]]
@@ -6628,6 +6629,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fb067e4cbd1ff067d1df46c9194b5de0e98efd2810bbc95c5d5e5f25a3231150"
 dependencies = [
  "cc",
+ "libc",
+]
+
+[[package]]
+name = "wait-timeout"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09ac3b126d3914f9849036f826e054cbabdc8519970b8998ddaf3b5bd3c65f11"
+dependencies = [
  "libc",
 ]
 
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -49,6 +49,9 @@ orion = "0.17"
 machine-uid = "0.6"
 base64 = "0.22"
 sysinfo = "0.33"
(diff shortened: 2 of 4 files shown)


yw-chan:
Thanks for the review — both points addressed in 446253d:

1. **gh process leak on `wait_timeout` `Err`**: `run_gh` now matches `Ok(Some) | Ok(None) | Err(_)` and kills+reaps the child on both timeout and wait error, so a hung gh can't linger.
2. **Windows explorer→terminal regression**: the effect no longer returns early on Windows. Only the cwd-reading poll + interval are skipped there; the `useWorkspaceStore` subscription (folder click → `cd`) stays active, since writing `cd` to the shell works fine on Windows.

Re-ran `tsc --noEmit`, `vitest run` (terminal suite 126 passed), and `cargo check` — all clean.

PR #158, opened 2026-07-08
fix: notification label follows a renamed tab

### LINKED ISSUE #157: Renaming a tab doesn't update the desktop notification's context label
## Summary

When a tab (a group of panes) is renamed, the desktop notification still shows the old auto-derived name (the cwd basename or transcript title), not the name the user gave the tab.

## Steps to reproduce

1. Enable "Track Claude session status" and desktop notifications.
2. Open a terminal tab running an agent, then rename the tab to a custom name (double-click the tab title).
3. Unfocus the window and let the agent finish or hit a permission prompt.
4. Observe the desktop notification.

## Expected

The notification's context label uses the tab's custom name, since the user explicitly renamed it.

## Actual

The notification keeps showing the cwd basename (or transcript title), ignoring the renamed tab title.

## Likely cause

`leafContextName` in `src/modules/claude-progress/lib/sessionNotifications.ts` resolves the label in this order:

1. transcript title from `titlesStore` (`titles[progressKey(cwd, agent)]`)
2. cwd basename (`basename(cwd)`)
3. `tab.title` — only when there is no cwd

Because a terminal pane almost always has a cwd, branch 2 wins and the user's renamed `tab.title` is never reached. A tab carries `renamed: true` once the user renames it (`tabsStore.setTabTitle`); that flag should take priority so a user-chosen name wins over the auto-derived cwd/transcript name.

## Notes

Same resolution path is shared by both approval and finished notifications, so both are affected.

mukiwu:
## Summary

Desktop notifications kept showing the auto-derived pane name (transcript title or cwd basename) even after the user renamed the tab, so a renamed group tab's notification never matched its new name.

`leafContextName` resolved the label as transcript title → cwd basename → tab title, and only reached the tab title when the pane had no cwd. A terminal pane almost always has a cwd, so the user's manual rename was never used.

## Fix

- Extract `resolvePaneLabel(tab, cwd, transcriptTitle)` and route it through `selectCardTitle`, the same helper the workspace card already uses. A renamed tab (`renamed: true`) now wins over the auto-derived name, so the toast matches what the card shows.
- Non-renamed panes keep the existing order: transcript title → cwd basename → tab title.
- Widen `selectCardTitle`'s first param to `Pick<Tab, "renamed" | "title">` since it only reads those two fields.

Both the approval and finished notifications share this path, so both are fixed.

## Test plan

- [x] `pnpm test` — 1347 passed, incl. new `resolvePaneLabel` cases (renamed-wins-over-cwd, transcript-over-cwd, cwd-basename fallback, tab-title fallback)
- [x] `pnpm typecheck` — clean
- [ ] Manual: rename a tab running an agent, unfocus the window, let the agent finish / hit a permission prompt, confirm the toast shows the renamed title

Fixes #157

### DIFF
--- a/src/modules/claude-progress/lib/sessionNotifications.test.ts
+++ b/src/modules/claude-progress/lib/sessionNotifications.test.ts
@@ -1,5 +1,39 @@
 import { describe, it, expect } from "vitest";
-import { notificationForTransition } from "./sessionNotifications";
+import { notificationForTransition, resolvePaneLabel } from "./sessionNotifications";
+
+describe("resolvePaneLabel", () => {
+  it("uses the tab's own title when the user renamed it, even with a cwd", () => {
+    const label = resolvePaneLabel(
+      { renamed: true, title: "My Group" },
+      "/Users/me/projects/api",
+      undefined,
+    );
+    expect(label).toBe("My Group");
+  });
+
+  it("prefers the transcript title over the cwd when the tab is not renamed", () => {
+    const label = resolvePaneLabel(
+      { renamed: false, title: "api" },
+      "/Users/me/projects/api",
+      "Fix auth bug",
+    );
+    expect(label).toBe("Fix auth bug");
+  });
+
+  it("falls back to the cwd basename when no transcript title is known", () => {
+    const label = resolvePaneLabel(
+      { renamed: false, title: "api" },
+      "/Users/me/projects/api",
+      undefined,
+    );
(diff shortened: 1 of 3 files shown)


mukiwu:
Addressed the medium finding in 7659f9c: gave the cwd-basename fallback test a distinct tab title (`Terminal`) so it no longer passes when the tab title is returned instead of the cwd basename.

PR #184, opened 2026-07-10
feat(context-menu): custom cut/copy/paste menu for text fields on Windows

yw-chan:
## Problem

On Windows the WebView2 default context menu (Reload / Save as / Inspect …) pops up on right-click across the app. It looks out of place in a desktop app, and its paste is slow (~5s) — the same problem `TerminalView` already works around with a custom menu.

## Changes

- **Windows-only**: suppress the WebView2 default context menu everywhere that isn't an editable field. macOS/Linux keep their richer native menus (the terminal deliberately relies on them there).
- Plain `<input>` / `<textarea>` fields get an app-styled **Cut / Copy / Paste / Select All** menu:
  - Paste reads through the fast Tauri clipboard path instead of the slow WebView2 one.
  - Password fields omit Cut/Copy.
  - Edits go through the native `value` setter + a dispatched `input` event (not the deprecated `execCommand`), so React-controlled inputs stay in sync.
- Rich editors (Monaco, Tiptap/ProseMirror) and the terminal keep their own menus — they're excluded by class/`defaultPrevented`.
- `ContextMenu` now flips at the viewport edge, so a menu opened near the bottom/right (e.g. the AI input at the foot of the panel) stays fully visible. Added optional `disabled` support to `ContextMenu` items (greyed Cut/Copy with no selection).
- New pure helpers in `src/components/inputMenuItems.ts` with unit tests.

## Test plan

- [x] `npx pnpm exec tsc --noEmit` — clean
- [x] `npx pnpm exec vitest run` — 1401 passed (12 new)
- [x] `npx pnpm tauri build` — NSIS + MSI bundles produced
- [x] Manual (Windows): right-click a bottom input (AI box) → custom menu opens **upward**, fully visible; Cut/Copy enabled only with a selection; Paste is instant and updates the field; password field shows only Paste/Select All; notes (Tiptap), editor (Monaco) and terminal keep their own menus; non-editable areas show no browser menu.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -45,6 +45,7 @@ import { invoke } from "@tauri-apps/api/core";
 import { useForwardStatusListener } from "@/modules/ssh/lib/useForwardStatus";
 import { sftpSessionStore } from "@/modules/ssh/lib/sftpSessionStore";
 import { enforceLogRetention } from "@/modules/logs/lib/sessionLog";
+import { InputContextMenu } from "@/components/InputContextMenu";
 
 const MIN_SIDEBAR = 180;
 const MAX_SIDEBAR = 640;
@@ -578,6 +579,7 @@ function App() {
       <UpdateToast />
       <NotifyToast />
       <SshPromptDialog />
+      <InputContextMenu />
       {pendingCloseAction && (
         <ConfirmDialog
           title={t("editor:closeUnsavedTitle")}
--- a/src/components/ContextMenu.tsx
+++ b/src/components/ContextMenu.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, type ComponentType } from "react";
+import { useEffect, useLayoutEffect, useRef, useState, type ComponentType } from "react";
 import { createPortal } from "react-dom";
 import type { LucideProps } from "lucide-react";
 import { useOverlayGuard } from "@/lib/overlayGuard";
@@ -11,6 +11,8 @@ export interface ContextMenuItem {
   onSelect: () => void;
   /** Render in the danger colour (used for destructive actions like Delete). */
   danger?: boolean;
+  /** Greyed and non-clickable (e.g. Copy with nothing selected). */
+  disabled?: boolean;
   /** Group index; a thin divider separates consecutive groups. */
   group?: number;
 }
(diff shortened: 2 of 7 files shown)


yw-chan:
Thanks for the review — addressed in 57d416a.

**Selection APIs throw on `number`/`email` (high)** — fixed at the root: `number` and `email` are removed from `TEXT_INPUT_TYPES`, so `isPlainTextField` no longer matches them and they fall through to plain browser-menu suppression + native `Ctrl+V` (this covers the SSH port fields in `ConnectionForm.tsx`). Added a defensive `getSelectionRange` helper and wrapped `setSelectionRange` in try/catch as belt-and-suspenders. New tests assert `number`/`email` are rejected and that `getSelectionRange`/`readFieldContext` never throw on them.

**Grouping (medium)** — kept the standard OS/browser layout (`Cut / Copy / Paste` │ `Select All`) rather than the suggested `Cut / Copy` │ `Paste / Select All`, and rewrote the comment to match. Grouping Paste with Select All reads as non-standard; separating Select All is the conventional split.

A parallel review pass also flagged two things I fixed here:
- **Cut atomicity** — Cut now deletes the selection only after the clipboard write resolves, so a rejected `writeText` can't drop the text with nothing on the clipboard.
- **Paste resilience** — Paste falls back to the web clipboard if the fast Tauri read fails, instead of silently no-op'ing.

**Accepted trade-offs (not changed):**
- *Undo stack* — edits go through the native value setter + `input` event instead of the deprecated `document.execCommand`, so Cut/Paste via this menu aren't on the browser undo stack. Re-introducing `execCommand` was explicitly out (deprecation), and there's no non-deprecated equivalent that feeds undo.
- *Native menu suppressed on non-text targets (images/links)* — this is the intended behavior on Windows: the feature exists to replace the out-of-place WebView2 menu. Open to preserving the native menu specifically on ``/`<a>` if that's preferred — happy to add it as a follow-up.

`tsc --noEmit` clean, `vitest run` 1404 passing.

mukiwu:
驗證結果：

- 本機（macOS）merge master 模擬無衝突，`tsc --noEmit` 乾淨，vitest 1489/1489 綠
- 確認 `IS_WINDOWS` gating 正確：macOS/Linux 完全不裝這個 handler，原生選單不受影響；`ContextMenu` 的視窗邊緣翻轉與 `disabled` 支援對所有平台的既有自訂選單是純改善
- Gemini review 的 selection API 例外問題（number/email input）確認已在後續 commit 全數處理，含防禦性 try/catch 與對應測試
- `actions.*` i18n key 路徑與既有結構吻合

合併。後續會以這個 PR 為基礎，把 macOS 的文字欄位選單也統一成同一套（另開 issue 規劃）。感謝貢獻！

PR #190, opened 2026-07-10
feat(menu): move macOS menu bar into the native system menu

### LINKED ISSUE #189: feat(macos): replace in-window menu bar with full native macOS menu
## Current behavior

On macOS the app shows **two menu bars**:

1. The system menu bar at the top with a minimal native menu (`TempoTerm` + `Edit` only, built in `src-tauri/src/modules/menu.rs`).
2. The self-drawn in-window `WindowMenuBar` (檔案 / 編輯 / 檢視 / 終端機 / 視窗 / 說明) rendered by `src/components/TitleBar.tsx` with menu definitions in `src/components/menuBarMenus.ts`.

This is redundant and looks off on macOS. See screenshot: the native bar says "TempoTerm  Edit" while the window title bar repeats a full menu underneath.

## Expected behavior

Do what VSCode does on macOS: put the **entire** application menu (File / Edit / View / Terminal / Window / Help) into the native macOS menu bar, and hide the in-window menu bar on macOS. Windows keeps the current in-window menu bar (it has no native menu by design).

## Implementation notes

- `src-tauri/src/modules/menu.rs` already builds the minimal native menu; extend it to the full menu tree currently defined in `src/components/menuBarMenus.ts` (Tauri 2 `MenuBuilder` / `SubmenuBuilder`).
- Menu items currently dispatch scoped `menu:*` events or call window actions from the frontend; native menu item handlers should emit the same events so `App.tsx` handling stays unchanged.
- Dynamic state (disabled items, checkmarks) needs syncing from frontend state to the native menu.
- Labels are localized in the frontend (zh/en); the native menu must receive the same localized strings, and update when the user switches language.
- Accelerators currently handled by the webview keydown handler in `App.tsx` should be declared on the native items so they show up in the menus (avoid double-firing).
- `WindowMenuBar` rendering becomes Windows/Linux-only (platform gate in `TitleBar.tsx`).

mukiwu:
Closes #189

## Summary

macOS previously showed two menu bars: the minimal native one (App + Edit) and the in-window `WindowMenuBar`. This PR moves the full menu (File / Edit / View / Terminal / Window / Help) into the native macOS menu bar, VSCode-style, and merges the tab row up into the traffic-light row. Windows is untouched: it keeps the custom title bar, the in-window menu, and the keydown shortcuts, and has zero behavior change.

## How it works

- `src/lib/nativeMenu.ts` serializes the existing `menuBarMenus.ts` tree (single source of truth) into a model: pre-translated labels, enabled flags, separators expanded from group boundaries, accelerators converted from mac glyphs to muda syntax, and predefined kinds for copy/paste/select-all (plus injected undo/redo/cut so system Cmd+Z/X routing survives).
- `src/lib/useMacNativeMenu.ts` (macOS-only hook) pushes the model via the new `set_native_menu` command on mount, language change, relevant store changes, maximize changes, and focus gain — microtask-coalesced, skip-if-unchanged, and only from the focused window since the macOS menu bar is app-global.
- `src-tauri/src/modules/menu.rs` rebuilds the NSMenu from the model (App submenu re-prepended on every rebuild since `set_menu` swaps the whole bar) and dispatches clicks via `on_menu_event` → `get_focused_window()` → `native-menu-click` event carrying the item id; the frontend routes it through the existing `executeMenuAction`, so all `menu:*` listeners in `App.tsx` are unchanged.
- `TitleBar` renders nothing on macOS; `TabBar` becomes the first row (its `pl-20` already reserved the traffic-light space from its pre-#178 top-row days).

Notable fix along the way: `AppHandle::set_menu` dispatches to the main thread internally — wrapping it in an explicit `run_on_main_thread` nested the dispatch and the NSApp bar never refreshed. The command now calls it directly (verified on-device).

## Behavior notes

- Menu accelerators are consumed by the system on macOS, so ⌥1–⌥7 now switch sidebar panels even while a text field is focused (same trade-off Chrome/VSCode make). The `App.tsx` keydown handler still serves Windows unchanged.
- Zoom-in shows as ⌘= in the menu (muda has no Plus token), matching Chrome/VSCode.
- Disabled items beep on their shortcut instead of silently no-oping.

## Test plan

- [x] `pnpm vitest run` — 1364 passed; the 7 collection failures (react-i18next mock missing `initReactI18next`) exist on base `e0f4817` and are unrelated (see follow-up below)
- [x] `tsc --noEmit` clean, `cargo test` 372 passed, `cargo build` zero warnings
- [x] On-device (macOS): six native menus render localized; language switch re-translates live; sidebar reorder updates the submenu + ⌥N hints; disabled items grey out; clicks dispatch to the focused window; multi-window focus re-pushes the model; Cmd+C/V/A/X/Z system routing intact
- [ ] Windows CI compile check rides the next `v*` tag build (workflow is tag-triggered; follow-up issue below proposes a per-PR check)
- Security/perf: `/tauri-check` reviewers both passed (0 critical/high)

## Follow-ups (will open issues)

- Per-PR `cargo check` for the Windows target so the cfg-split command stays warning-free before release tags
- Pre-existing: 7 vitest collection failures from the shared react-i18next mock; pre-existing clippy debt in `pty/session.rs` and `sessions_index/*`


mukiwu:
Addressed all five review comments in 583d2e0: `focused` now initializes to `false` so a background window can never overwrite the app-global menu before its focus state resolves (the high-priority race), catch handlers added to `isFocused`/`isMaximized`/unlisten promises, and the App submenu title now comes from `package_info().name` instead of a hardcoded string. Hook suite 6/6, tsc clean, cargo test 372 green.

PR #193, opened 2026-07-10
test: restore localStorage in Node 26

saamuelng601-pixel:
## Summary
- restore jsdom localStorage in the Vitest setup when Node 26 exposes a disabled global accessor
- add a regression test that verifies the test global uses jsdom storage

## Validation
- `pnpm typecheck`
- `pnpm test`: 1,337 tests pass; this removes the localStorage cascade. The remaining 7 failed suites are the pre-existing `react-i18next` incomplete-mock issue, intentionally out of scope for this focused test-harness fix.

### DIFF
--- a/src/test/setup.test.ts
+++ b/src/test/setup.test.ts
@@ -1,0 +1,11 @@
+import { describe, expect, it } from "vitest";
+
+describe("test setup", () => {
+  it("exposes jsdom localStorage on the test global", () => {
+    const testDom = (globalThis as typeof globalThis & {
+      jsdom: { window: Window };
+    }).jsdom;
+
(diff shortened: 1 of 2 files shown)


mukiwu:
Verified twice against real Node 26 (v26.5.0) before merging:

- Reproduced the failure on master: the disabled `localStorage` accessor Node 26 exposes makes 7 of 12 `uiStore` tests fail, exactly as described.
- This branch merged with today's master: full suite green on Node 22 (1521/1521, including the new regression test) and the localStorage-dependent suites green on Node 26 (39/39), while master alone still fails under Node 26.

The repo pins Node 22 in CI, so this only bites contributors on newer Node — but Node 26 is current and this is a clean, test-harness-only fix. Thanks for the contribution!

PR #222, opened 2026-07-16
feat(worktrees): add the manager modal and its status-bar badge

mukiwu:
Third slice of Parallel Worktrees, on top of #220 (Rust commands) and #221 (stores). This is the first part you can actually see.

## What lands

- **Status-bar badge** — counts every worktree across every repo known to have one; hidden at zero, like Ports. Clicking opens the manager on everything.
- **Manager modal** — one surface, two scopes. The entry point picks which, so nobody lands in the wrong one: the badge means "everything I have lying around", a terminal (later slice) means "the repo I am working in".
- **Rows** — branch, path, uncommitted count, locked / stale, and a dot for whichever agent is busy in that directory.

## Why the badge is ambient

Creating a worktree will be one click, and each one carries its own `node_modules`. Without a number sitting in view they accumulate invisibly until the disk is full. The badge reads cached per-repo counts from the registry and never scans to render.

## The discovery loop this had to break

The badge reads the registry; the registry was only filled by `refresh()`; `refresh()` was only called by the modal; the modal was only reachable from the badge — which hides while the registry is empty. A feature that could never appear.

`useWorktreeDiscovery` breaks it by watching `useWorktreeStore.infos`, which the workspace cards already populate with `git_worktree_info` for every open terminal's cwd. That answer carries the repo's main path, so discovering a repo costs **zero extra IPC**. Each repo is scanned once, on an event, not a timer.

## Deliberately not here

Open / split / remove — each has its own slice. Rows are read-only. Dirty counts and disk size are lazy per row: measuring walks the whole checkout, so it only ever happens because someone asked for that row.

## Testing

- `pnpm typecheck` clean; 43 worktree tests pass, full suite green apart from two TitleBar hover-delay failures that reproduce on `master` (pre-existing, unrelated).
- Ran the real app: three columns fine, no crash, and the badge correctly stays hidden — no open terminal was inside a repo with worktrees.
- To see it appear: open a terminal in a repo that has a worktree, and the badge shows up bottom-right next to Ports.


mukiwu:
Thanks — all three land, and the first one was a real bug rather than a theoretical one.

**Stale dirty count (high)** — confirmed by writing the test before the fix: a row that goes prunable keeps its component instance, so the count loaded while the directory still existed stayed on screen right next to the warning saying it's gone. Fixed, and the test now pins it.

**Measure button on a stale row (medium)** — agreed, a disabled control next to "directory is gone" adds nothing the warning doesn't already say. Omitted entirely.

**Duplicated `formatBytes` (medium)** — taken, with one correction: the two weren't identical. The local copy rendered `512 MB` where sysmon's renders `512.0 MB`. That difference was gratuitous — PortRow and StatusBar sit on the same status bar and already use sysmon's, and it's the one with tests. Now reused.

PR #223, opened 2026-07-16
feat(worktrees): open a worktree, or split it beside your work

mukiwu:
Fourth slice of Parallel Worktrees, on top of #222. The manager could only look; now it can take you somewhere — which is the point where the feature stops being a dashboard.

## What lands

- **Open** — the row's body is the target, so the way in is the whole line rather than a control you have to find. Spawns a terminal in that worktree, in a tab of its own.
- **Split** — puts the worktree beside the pane you are in. This is what you want when you are comparing two worktrees rather than moving to one.
- **Focus instead of duplicate** — opening a worktree that is already open takes you to the shell that is there. Two terminals in one worktree is a thing to ask for, not a thing to get by accident.

## Three traps this had to avoid

**Never `setRoot`.** Pointing the explorer at a directory types `cd` into whatever shell has focus — straight into a live agent's prompt. Both paths go through the tab store instead. The file explorer's own context menu already carries this warning in a comment; same rule here.

**The cwd goes on the pane, not the tab.** `resolveTerminalCwd` is `paneCwd || rootPath || tabCwd`, so a tab carrying only `cwd` spawns its shell wherever the *explorer* happens to be pointing. `newTerminalTab` seeds both; the split passes the cwd in its pane content. A hand-rolled tab that set only `Tab.cwd` would be a live bug.

**Prefix matching would open the wrong worktree.** Worktrees are siblings in `<repo>-worktrees/`, so `startsWith` calls `feat-ab` a match for `feat-a`. `findWorktreePane` uses `isUnder`, whose boundary falls on a separator. A pane that has cd'd deeper still counts — `cd src` does not leave the worktree.

## Edges covered

- Stale (`prunable`) and bare worktrees cannot be opened — there is no directory to spawn a shell in. The row says so instead of failing at the pty.
- Split is offered only when there is a pane to sit beside, and stops at the pane cap. `splitPaneWith` takes an explicit target rather than the active pane, so unlike `splitActivePane` it does not enforce the cap for itself — this caller does.
- That cap was two bare `>= 8`s in `tabsStore`. This is the third place that has to agree on it, so it is now `MAX_PANES`.

## Testing

- 9 tests on `findWorktreePane` (sibling worktrees, the parent repo, cd'd-deeper panes, splits, non-terminal panes, Windows case-folding), 7 driving the row against the real tab store — asserting on where terminals actually landed rather than on spies.
- `pnpm typecheck` clean. Full suite: 1630 pass; the two TitleBar hover-delay failures reproduce on `master` (pre-existing, unrelated).
- No `src-tauri/` changes.

### DIFF
--- a/src/i18n/locales/en/worktrees.json
+++ b/src/i18n/locales/en/worktrees.json
@@ -10,14 +10,20 @@
   "row": {
     "main": "main",
     "detached": "detached",
-    "open": "Open in this window",
+    "open": "Open",
     "dirty": "{{count}} uncommitted change",
     "dirty_other": "{{count}} uncommitted changes",
     "locked": "Locked",
     "lockedReason": "Locked: {{reason}}",
     "stale": "Directory is gone",
     "measure": "Measure size",
-    "measuring": "Measuring…"
+    "measuring": "Measuring…",
+    "openHint": "Open a terminal in this worktree",
+    "focusHint": "Go to the terminal already in this worktree",
+    "split": "Split",
+    "splitHint": "Open it beside what you are working in",
+    "splitFull": "This tab is full",
+    "openStale": "The directory is gone — nothing to open"
   },
   "agent": {
     "active": "Working",
--- a/src/i18n/locales/zh-Hant/worktrees.json
+++ b/src/i18n/locales/zh-Hant/worktrees.json
@@ -10,14 +10,20 @@
   "row": {
     "main": "主要",
     "detached": "分離 HEAD",
-    "open": "已在這個視窗開著",
+    "open": "開啟",
     "dirty": "{{count}} 個未提交變更",
     "dirty_other": "{{count}} 個未提交變更",
     "locked": "已鎖定",
     "lockedReason": "已鎖定：{{reason}}",
     "stale": "目錄已不存在",
     "measure": "計算佔用",
-    "measuring": "計算中…"
+    "measuring": "計算中…",
+    "openHint": "在這個 worktree 開一個終端機",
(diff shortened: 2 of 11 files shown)


mukiwu:
All three land. The first one was worse than described — thanks for it.

**SSH panes (high)** — real, and the trigger is reachable: `ssh` is a flag on ordinary terminal content, so a `kind === "terminal"` check lets remote shells straight through. An SSH pane with no cwd of its own inherits its tab's, and `openFromSidebar` splits SSH panes into whatever tab is active — including one already sitting in a worktree. Opening that worktree would focus a shell on another host and call it yours.

The same scan is also in `worktreeSessionStatus`, **already on master**, where it reports a remote host's agent as that worktree's — the same misattribution class caught in #221. Wrote a failing test against each before fixing; both were red, so neither was theoretical.

Rather than adding an `ssh` check in two places, both now call `localTerminalCwd`, which answers "where is this pane's shell on *this* machine" once. The rule has one home.

**Launcher tabs (medium)** — correct, and worth stating plainly: `TabsArea` renders `LauncherPanel` for those and never reads their pane tree, so the split would have silently vanished. Split is no longer offered there. Open still is, since "a tab of its own" is well-defined from anywhere.

**Guarding the handler (medium)** — taken, with a change. Re-checking the captured `activeTab` inside the handler would re-check a stale value: it is this render's snapshot, and which tab is active can move between paint and click. The handler now re-reads the store and guards the fresh tab, which covers the raced case you were pointing at rather than only the clicked one.

Suite: 66 worktree tests, 1640 overall. The two TitleBar failures reproduce on `master`.

PR #224, opened 2026-07-16
feat(worktrees): create one by naming a branch

mukiwu:
Fifth slice, on top of #223. Until now the manager could only show worktrees that already existed — made at a CLI, by someone who already knew about `git worktree`. Now it makes them.

## What lands

Name a branch → see exactly where it will land → create → a terminal opens inside it.

## On reachability

The create flow's planned entry point is a terminal's `⋯` menu, which is a later slice. Building it behind an entry point that does not exist yet is precisely how the badge in #222 nearly shipped invisible, so it is reachable today instead:

- **Repo scope** names its repo in the header, so a single **New** button is unambiguous.
- **Global scope** (what the badge opens) has several repos, so each group header carries its own **New** — the header already says which repo.

Two tests pin that, because "it's reachable" is a claim, not a fact.

## Decisions worth reviewing

**The path is shown before anything exists.** A worktree is a real directory that will carry its own `node_modules`. Being surprised by where it went is something you find out much later, from a full disk.

**Branch names are validated here as well as in Rust.** This is the friendly half — it tells you while you type. It is not the only half: the Rust side refuses the same names, and when git knows something this side cannot (existing branch, occupied directory, repo mid-rebase), git's own message is shown verbatim rather than paraphrased.

**The slug flattens rather than nests.** `feat/x` becomes one directory, not `feat/` holding `x`, so removal leaves no empty parents. That trades for a collision — `feat/x` and `feat-x` want the same directory — which the Rust side already catches by refusing a non-empty target. Loud, not silent.

**The terminal opens at the path git reports**, not the one we computed; canonicalization can make them differ.

## Deliberately not here

The after-create hooks — setup command (`pnpm install`), copying gitignored `.env*`, launching an agent — are the next slice. The per-repo settings store already carries their fields (#221); nothing reads them yet. This slice is create itself, which is reviewable on its own.

## Testing

- 17 tests on path resolution and branch validation (Windows separators, container override, every git ref rule, flag-shaped names).
- 3 on the store's `create` (rescans on success so the badge counts it; does not rescan on failure — nothing new to find).
- 7 driving the form: live path preview, container override, validation while typing, git's error surfaced, no double submit.
- 2 on reachability from each scope.
- `pnpm typecheck` clean. Full suite 1669 pass; the two TitleBar failures reproduce on `master`.
- No `src-tauri/` changes — `git_worktree_add` shipped in #220.


mukiwu:
Both real, both taken — the first one especially. Wrote a failing test for each before fixing; both were red.

**Rescan failure rejecting create (high)** — the worst kind of wrong: the worktree *is* on disk, and we tell the user it isn't. Their retry then dies on "branch already exists", using the evidence of success to argue it failed. The list catches up on the next scan; the lie would not.

Took the try/catch without the `console.error` — a scan losing a race with a git lock is expected, not exceptional, and it is already swallowed the same way where the modal refreshes. The comment carries the reasoning instead.

**Windows drive root (medium)** — correct, and worth naming precisely: `C:` is not a shorter spelling of `C:\`, it is drive-*relative*, so it would be both refused by the Rust side for not being absolute and wrong if it weren't. Unix's `/` was already handled — this was the asymmetry, not a missing edge case. Fixed with tests for both roots.

The two test suggestions are folded into the above rather than added separately.

PR #225, opened 2026-07-16
feat(worktrees): make a worktree usable, removable, and findable

mukiwu:
The last three slices of Parallel Worktrees, combined as one on request. On top of #224.

---

## Usable — the after-create hooks

`git worktree add` checks out tracked source only. A fresh worktree has no `.env` and no `node_modules`, so an agent's first command dies on both. Three options fix that: copy the gitignored files you name (default `**/.env*`), run a per-repo setup command, start an agent.

The copy runs **before** the terminal that will read those files starts, and setup + agent go in as one `&&` line — an agent that starts in a worktree where `pnpm install` just failed is worse than no agent, because it works against a broken tree while the failure scrolls away above it.

### Which files exist is git's answer, not ours

`git status --porcelain -z --ignored` lists ignored **files** individually but collapses an ignored **directory** into one entry it never descends into. That is exactly the rule this needs — `node_modules/foo/.env` is not the user's file, and copying it would conjure a dependency tree in a worktree that has none.

I first wrote the walk by hand with the `ignore` crate. The security review caught that it silently disagrees with git the moment a repo declares `node_modules/` in `packages/app/.gitignore` rather than at the root — **and that my test could not see it**, because the fixture only put `node_modules/` in the root ignore file. Verified against real repos with `git check-ignore` before changing anything. The fixture now declares one ignored dir at each level.

### No initial prompt for the agent — deliberate

The agreed design had one. Passing it means quoting free text for a shell we do not know: the pty runs a **login shell the user can override**, and on Windows that is PowerShell *or* cmd.exe, which escape quotes differently and share no safe form for arbitrary text. A prompt is one paste away once the agent is up. A mangled one is a silent wrong answer. Everything else in that slice landed.

---

## Removable — the only place this can destroy work

- Uncommitted work **blocks**, and says how much rather than that there is some. `force` reaches git only from a ticked box — never a default, never inferred. A test asserts the exact argument.
- A count that has not loaded counts as dirty. Not knowing is not evidence there is nothing to lose.
- The branch is **kept** unless asked for; force-deleting an unmerged one is a second box.
- The repo's own working tree is not a confirmation away — it is a no.
- **Every terminal in the worktree is closed before git touches the directory.** On Windows a live pty holds its cwd open and the removal fails halfway, leaving git pointing at a directory that is half gone. macOS unlinks happily, which is why this only breaks on someone else's machine — hence a test on the ordering.
- Prune, for entries whose directory is already gone.

---

## Findable — the entry points

The badge cannot introduce this feature; it hides until a worktree exists. So:

- **A `⋯` on any terminal pane inside a git repo**, with the one-time hint. Local panes only — a worktree is a directory on *this* machine, and an SSH pane's shell is on another.
- **A branch in the git graph** offers a worktree of its own: unlike checkout, it leaves the current tree and whatever is running in it alone.

**No keyboard shortcut.** It would need hand-syncing across three lists (the real handler, the menu-bar hints, the settings help); wired into one of three is a lie. `⌘W` is "close" everywhere and not mine to take. The two entry points above cover discovery.

---

## Drive-by: a live bug in the git graph

Its worktree picker called `setRoot`, which drives cwd sync — **typing `cd <path>` into whatever shell has focus**, quite often a running agent's prompt. It opens a terminal instead, and reuses the one already there.

Two existing tests pinned that behavior as correct ("switches the workspace root", "posts a file-explorer-updated notice"). They assert what should happen now. The toast said "File explorer updated" — describing the side effect that no longer occurs — so its string is gone too.

---

## Reviews run

- **Security** (`/tauri-review`): 2 HIGH, both fixed and both verified failing first. Empty globs failed open and copied the **entire repo** (an empty override set reports every path as unfiltered, and `settings.copyGlobs ?? DEFAULT` never fires for `[]` — a cleared field). Nested `.gitignore` above. Plus: a worktree inside the repo made the copy find its own output; the lexical `starts_with` guard protected nothing it claimed to; the `..` glob rule had false positives (`.env..bak`) and guarded an escape that is not reachable — that test now asserts the property rather than my guard.
- **Performance** (`/tauri-perf`): PASS. It independently confirmed the directory-collapse behavior this design rests on (3,000-file `node_modules` → one entry, ~15ms).

## Testing

- Rust: 422 pass, `RUSTFLAGS="-D warnings" cargo check` clean. 8 on `copy_local_files` alone, against real repos.
- Frontend: 1709 pass. The two TitleBar hover-delay failures reproduce on `master`.


mukiwu:
Right, and worth naming plainly: this is the same bug I fixed in #224, reintroduced in a different shape.

There, a rescan that lost a race with a git lock rejected the whole create. Here, a failed copy holds the form open. Both end the same way — the worktree is on disk, the form is still up, and the only thing left to click is Create, which dies on "branch already exists". **Using the evidence of success to argue it failed.** I wrote the comment about that trap and then walked into it fifteen minutes later.

Taken as suggested: notify, then `onDone()`. The failure still gets said out loud rather than swallowed, because an agent that needs `.env` will die without it — but it is said on the way out, not as a wall.

Test rewritten to assert the way out (notice posted, form closed, terminal open) rather than the message on the wall. It was red first.

PR #240, opened 2026-07-18
chore(dock): remove the agent-status dots from the strip icons

mukiwu:
## Summary

Removes the v0.2.0 aggregate agent-status badges (the small spinning/pulsing dots) from the sidebar strip icons of the workspaces and sessions panels, per maintainer decision.

The whole mechanism goes, not just the flags: `StatusStripDot` (+ its tests), the `showSessionStatus` registry field, the `DockIcon` status subscription, and the now-unconsumed `selectSessionStatus` / `aggregateSessionStatus` helpers. `AGGREGATE_PRIORITY` stays — workspace cards and worktree rows still reduce per-leaf statuses through it, and their own status indicators are untouched.

## Tests

- Frontend: 1772 passed across 201 files (the removed component's 6 tests go with it); `tsc --noEmit` clean.


Katomoto (commented):
Approach is solid. Minor nit: this would benefit from a brief comment explaining why we skip validation for internal callers.

PR #293, opened 2026-07-25
fix(workspace): resolve Codex card titles per session, not per directory

dca:
## Problem

With several Codex sessions running in one directory, every workspace card showed the same title — whichever session wrote to its rollout last renamed all of its same-cwd siblings. Claude sessions were already correct (fixed per-session in #233); Codex still cross-linked.

## Root cause

The title pipeline already carries a per-pane Codex session id (the status hook reports it; the resume path in #260 relies on it) and the title cache already keys per session — but `fetchTitle` dropped the id at the bridge, and the backend `codex_session_title(cwd)` answered with the directory's newest rollout. Per-session cache keys, one shared per-directory answer.

## Fix

- Thread the session id through `fetchTitle` → `codexSessionTitle` → the `codex_session_title` command (mirrors the `claude_session_title` shape).
- `RolloutCandidate` now records the rollout's `session_meta` `payload.id` as `rollout_id` (parsed in the same single first-line pass as the cwd).
- A provided id resolves by exact match on `payload.id` only — never `payload.session_id`, which on subagent rollouts names the *root* session while living in the same cwd (matching it would dress a session in its own subagent's prompt), and which pre-2026-07 rollouts don't have at all.
- The exact match is global (no cwd filter: ids are unique, and a `/tmp` vs `/private/tmp` mismatch must not blank a hit) and never falls back to newest — a just-started titleless session must not wear a concurrent sibling's title. Only an id-less request keeps the legacy newest-for-cwd behavior (SSH remotes, panes with no reported id).

## Review follow-ups (from the first review pass)

- Rebased onto latest master; the changelog entries now sit under the fresh post-v0.3.2 template.
- Updated the five comments that still described Codex as staying on the legacy per-directory key, and made the `codexSessionTitle` doc state the no-fallback rule precisely.
- Renamed `RolloutCandidate.id` → `rollout_id` with a doc comment distinguishing it from `payload.session_id`.
- Test helper `write_rollout` now takes a named-field `RolloutSpec` (the two adjacent same-typed id params can no longer be swapped silently), and a `TempBase` RAII guard cleans the temp dir even when an assert fails.
- Collapsed `fetchTitle`'s agent branch into a function pick, since both bridges now share one signature.

## Tests

[...]

PR #305, opened 2026-07-30
feat(settings): add custom background images

mark22013333:
## Summary

Adds app-managed custom background images to **Settings → Appearance**, with controls designed for readable terminal and workspace surfaces.

## User experience

- Choose or drag-and-drop a PNG, JPEG, or WebP image to add or replace the background.
- Preview the image at 16:9 before using it.
- Choose **Workspace** or **Whole window** scope.
- Adjust interface image opacity and terminal image opacity independently.
- Override interface and terminal default text colour while preserving ANSI colours.
- Remove the managed image without touching the original source file.
- Keep settings, confirmation, and security dialogs on solid surfaces for readability.
- Preserve Claude Code colour output even when TempoTerm is launched by a parent process with `NO_COLOR`.

## Storage and security

- Supports PNG, JPEG, and WebP only, with a 20 MiB limit.
- Validates the extension, file signature, regular-file type, and actual bytes read.
- Copies the selected image into the app-managed `appearance` directory using a temporary file and atomic replacement.
- Replacement failures preserve the current background.
- Removal is restricted to the managed background directory and accepts no arbitrary path.
- Does not relax CSP, asset protocol permissions, network access, or Tauri capabilities.
- Stores only the managed path and appearance settings in Zustand; image bytes are never written to localStorage or encoded as base64.

## Terminal rendering

- Enables xterm transparency while keeping the existing ANSI palette.
- Applies the wallpaper tint once across the terminal header, gutter, scrollbar edge, and xterm surface.
- Avoids compounded masks in nested workspace and whole-window surfaces.
- Keeps terminal image visibility separately adjustable so wallpaper detail does not require weakening the rest of the interface.

## Validation

- `pnpm typecheck` — passed
- Frontend suite — 1,965 tests passed; one existing Git Graph async timing test was flaky in the full run and passed 14/14 when rerun in isolation
- `pnpm build` — passed
- Rust suite — 475 tests passed in the sandbox; the one localhost-port test blocked by sandbox permissions passed when rerun with local network access; 6 environment-dependent tests ignored
- `cargo check --manifest-path src-tauri/Cargo.toml` — passed
- `cargo fmt --manifest-path src-tauri/Cargo.toml --check` — reports repository-wide pre-existing formatting drift, including files outside this PR
- Manual validation — dark/light themes, workspace/whole-window scope, terminal transparency, Claude Code ANSI colours, drag replacement, custom text colour, and pane surface consistency

## Screenshots

| Theme | Workspace | Whole window |
| --- | --- | --- |
| Dark | Dark workspace | Dark whole window |
| Light | Light workspace | Light whole window |


mukiwu:
Thanks for this — it's a substantial piece of work, and the security side came back clean. I ran the diff through four separate review passes (standards, spec, Windows/cross-platform, and a dedicated Tauri security review) and verified the findings against the code myself. Summary below, split into what blocks merge and what's open for discussion.

First, credit where it's due. The security review found **0 critical / 0 high / 0 medium** issues. That's not a given for a feature that copies user-chosen files into app data. Three things in particular are done right:

- The destination path is decided entirely in Rust (`unique_filename()` + `app_data_dir()`). The frontend's `source_path` is only ever a read source and never participates in building the write path, so path traversal doesn't apply structurally rather than being filtered out.
- `appearance_remove_background_image` takes no arguments at all.
- Format validation runs on an already-open file handle and the copy seeks back to 0 on that *same* handle, so validation and copy can't be raced apart. The size cap is re-checked against actual bytes copied, not just `metadata.len()`.

The Rust test coverage (header/extension mismatch, directory-as-file, copy limit, failed-replacement preservation, unmanaged-file safety) covers real security paths, not just the happy path.

---

## Blocking

**1. The branch conflicts with master and CI has never run.**

`mergeable: CONFLICTING`, and `gh pr checks` reports `no checks reported` for this branch. Please rebase onto current master and push so `windows-check.yml` and the frontend check actually run. Everything under "Windows" below is unverified precisely because CI has never executed here.

**2. Please split the `NO_COLOR` change into its own PR.**

`src-tauri/src/modules/pty/session.rs:63-65,90` — `strip_parent_color_opt_out()` has nothing to do with background images, and both commits are scoped `feat(settings)` / `fix(settings)`. Beyond the scoping, it needs changes before it lands anywhere:

- **The test doesn't cover the case it claims to fix.** `session.rs:499` calls `cmd.env("NO_COLOR", "1")` explicitly and then removes it. The bug being fixed is an *inherited* value from the parent process; that path is never exercised.
- **On Windows it overrides a deliberate user preference.** `CommandBuilder::new` seeds from `get_base_env()`, which on Windows additionally reads `HKLM\...\Session Manager\Environment` and `HKCU\Environment`. A `NO_COLOR` set through System Properties → Environment Variables lands in exactly that key — that's the user's persistent, intentional setting, not a leaked implementation detail from a parent tool.
- **The escape hatch in the comment doesn't exist on Windows.** The comment says "a user's login-shell config still runs afterwards and can deliberately set NO_COLOR again", but `login_args` (`src-tauri/src/modules/pty/shell.rs:229-235`) only returns `-l` for zsh, bash, and fish. PowerShell still auto-runs `$PROFILE`, but **cmd.exe users have no recovery path at all**.

A cheap discriminator if you want to keep the behaviour: strip only when `cmd.get_env("NO_COLOR").is_some()` *and* `std::env::var_os("NO_COLOR").is_none()` — that combination can only mean it came from the registry, i.e. the user set it deliberately. Otherwise gate the strip behind `!cfg!(windows)`.

Also note `NO_COLOR` is a documented accessibility opt-out, so silently overriding it deserves a setting rather than being unconditional.

**3. The 16:9 preview described in the PR body isn't implemented.**

> "Preview the image at 16:9 before using it."

`BackgroundImageSettings.tsx:63-73` — `chooseImage()` calls `saveImage()` immediately after the file dialog resolves, which copies into app data, updates the store, and deletes the previous managed file. The 16:9 box at `:178-189` renders the image that has *already been committed*. A mis-click therefore destroys the existing background with no undo. Either add a real pre-commit preview step or adjust the description.

**4. 1.3 MB of unreferenced PNGs.**

`screenshots/custom-background/*.png` — four files, nothing in the repo references them (`grep -rn custom-background --include="*.md" .` finds nothing), and the PR body links them from your fork's `raw.githubusercontent` URLs rather than these paths. Existing files under `screenshots/` are all README-linked. This repo has an explicit keep-the-installer-light constraint, so please either wire them into the README or drop them from the branch and keep them in the PR description only.

**5. Changelog entries are missing PR numbers.**

`CHANGELOG-NEXT.md` — every entry in `docs/changelogs/CHANGELOG-0.3.3.md` ends with ` (#300)` / ` (#293)`. Both new entries need ` (#305)`. The `fix` entry also currently bundles the `NO_COLOR` change, which follows from item 2.

---

## Windows-specific (unverified — CI has never run)

**Deletion fails hard on a locked file.** `src-tauri/src/modules/appearance/mod.rs:135` — `fs::remove_file(path).map_err(|_| AppearanceError::StorageUnavailable)?`. On Windows, unlink is `DeleteFileW`, which returns `ERROR_SHARING_VIOLATION` if any process holds a handle without `FILE_SHARE_DELETE` — Defender's real-time scanner routinely does this to a just-written file. macOS unlinks regardless, so this never reproduces locally.

In the save path the call site uses `let _ =` and degrades to a stale file, which is fine. But `appearance_remove_background_image` (`:236`) propagates it, so "Remove" will intermittently surface the generic error even though the image disappears on retry. Compounding it: `BackgroundImageSettings.tsx` calls `clearBackgroundImage()` only *after* the invoke resolves, so the `` is still mounted and its asset request may be in flight while the delete runs.

Suggested fix: treat `PermissionDenied` / os error 32 like the existing `NotFound` arm, or retry ~3× with a short backoff; and unmount the preview before invoking removal.

**DPI handling in `nativeDragCoordinates.ts`.** The helper accepts both the raw point and the `/devicePixelRatio` point. Tauri v2 documents `onDragDropEvent`'s position as `PhysicalPosition` on every platform, so the "some platforms report logical coordinates" premise doesn't hold, and accepting both makes the drop target claim two disjoint screen regions. Windows is where this bites: 125%/150% scaling is the laptop default and WebView2's `devicePixelRatio` tracks the window's current monitor, so a mixed-DPI setup desyncs mid-drag. Suggest using `x / devicePixelRatio` only.

---

## Correctness, non-blocking

- **`flush()` without `sync_all()`** — `appearance/mod.rs:189-197`. The rename is genuinely atomic (same directory, same filesystem) and failure does preserve the current image, but without an fsync before the rename a crash can leave the committed file truncated.
- **Validation errors never reach the user** — `BackgroundImageSettings.tsx:56-58`. The `catch {}` discards the typed `AppearanceError` and substitutes one generic string, so `UnsupportedFormat`, `FileTooLarge`, and `InvalidFile` are indistinguishable to the user. Worth mapping them to distinct i18n keys, since the Rust `Display` strings are English-only and can't be surfaced directly. (Those strings also say "20 MB" where the limit is 20 MiB.)
- **`is_managed_background` accepts `.tmp`** — `appearance/mod.rs:114`. Post-commit cleanup can therefore delete another in-flight import's temp file, and that import's rename then fails. The `busy` guard makes it unlikely within one window, but this app has `open_new_window`, so two windows importing at once can collide. Same root cause makes the second window's committed file disappear from under the first.
- **`BackgroundImageLayer` has no `onError`** — unlike the settings preview. `backgroundImagePath` persists in localStorage, so a managed file deleted out from under the app leaves a permanently broken layer.
- **Short read on the header** — `appearance/mod.rs:159` uses a single `read()` for 16 bytes. `Read::read` is allowed to return fewer bytes than requested; WebP needs 12. Unlikely for a regular file, but `read_exact` with a short-file fallback is sturdier.
- **`hexToRgba` only matches `#rrggbb`** — `src/modules/terminal/lib/backgroundTheme.ts:5`. Every theme in the repo currently uses that form, so this isn't live today, but any other notation would silently return opaque and make the wallpaper vanish behind the terminal.

## Conventions

- **Component naming** — `src/modules/settings/BackgroundImageSettings.tsx` should be `BackgroundImageSettingsSection.tsx` (plus its test). All six siblings follow `*SettingsSection.tsx`.
- **Cross-module DOM coupling** — `TerminalView.tsx:1465` does `document.querySelector("[data-native-file-drop-zone]")` to learn about a settings-pane element. It works today because exactly one element sets that attribute, but the selector is unscoped, so any future drop zone anywhere silently disables terminal file-drop. A shared registry in `nativeDragCoordinates.ts` would be sturdier.
- **Repeated appearance tuple** — `backgroundImagePath` + opacity + scope + textColor are read together in `App.tsx:158`, `DockShell.tsx:40`, `BackgroundImageLayer.tsx`, `PaneTabContent.tsx:90`, and `TerminalView.tsx:217`, and the active-predicate `Boolean(path) && opacity > 0 && scope === X` is duplicated verbatim in the first two. A `useBackgroundImage(scope)` hook would collapse all five.
- **Duplicated drag plumbing** — the `onDragDropEvent` subscribe/dispose lifecycle is now written twice (`BackgroundImageSettings.tsx:96-149` and `TerminalView.tsx:1472`). `nativePointInElement` was extracted; the subscription wasn't.
- **`index.css`** overrides the solid tokens twice, via `.wallpaper-solid-surface` and `.wallpaper-surface [role="dialog"][aria-modal="true"]`, with identical bodies. One mechanism would do.
- **Test literal** — `PaneTabContent.test.tsx` asserts `rgba(34, 34, 34, 0.685)`; deriving it from `backgroundSurfaceAlphas` avoids hand-recomputing when a factor changes.

## Low-severity security notes

None of these are live holes; recording them for completeness.

- Residual TOCTOU between `fs::metadata(&source_path)` and `File::open(&source_path)` (`mod.rs:149-157`). Size and format are both re-checked on the handle, so a swap can't smuggle a non-image through; the remaining exposure is a swap to a FIFO blocking the worker thread. Opening first and taking metadata from the handle would close it.
- The command can copy any readable file into a directory the asset protocol allows, which in principle routes around the `assetProtocol.scope.deny` list. In practice the source must be a genuine PNG/JPEG/WebP, and the app already exposes `fs_read_file`, so the marginal gain is nil.
- `settingsStore`'s zustand `persist` has no `merge`/`migrate`, so values restored from localStorage bypass `clampBackgroundImageOpacity` and `normalizeBackgroundImageTextColor`. Adding those to a `merge` would also stop a corrupted localStorage from breaking the UI.

---

Happy to look again once the rebase lands and CI is green. The feature itself is well-built — most of the above is scoping and platform coverage rather than the core design.

mark22013333:
Thanks for the review — I’ve pushed a follow-up that addresses the feedback and tightens the feature scope.

- Removed the unrelated `NO_COLOR` behavior and the committed preview screenshots
- Added staged background settings with a live, interactive workspace preview; nothing is persisted until **Apply**
- Fixed native image drag-and-drop ownership, macOS Retina coordinate handling, and terminal path leakage while Settings is open
- Hardened app-managed image saving/removal, including structured errors and Windows-safe file replacement retries
- Fixed split-pane header/chrome masking for Git Graph and other non-terminal panes
- Updated `CHANGELOG-NEXT.md` with `(#305)`

Validation completed locally:

- `pnpm typecheck`
- `pnpm build`
- 222 test files / 1,975 tests passing

Ready for another look when convenient. Thank you!

mukiwu:
waiting for my weekly token reset... 😂

mark22013333:
Got it, no worries at all

mukiwu:
合併了，感謝你這一輪的回應，改得很徹底。

先講兩件跟你分支有關的事：

**CI 一直沒跑不是你的問題。** GitHub 對第一次投稿的貢獻者會要求 maintainer 手動批准才跑 workflow，你 7/30 建立的那兩個 run 從頭到尾卡在 `action_required`，沒人去按。我上一輪說「要 rebase 才會跑 CI」是錯的，rebase 只解衝突，跟這個無關。我已經批准了，兩顆都綠：

```
cargo-check-windows   pass  1m25s
typecheck-and-test    pass  2m14s
```

`cargo-check-windows` 這顆特別有意義，它跑的是 warnings-as-errors 的 `cargo check`，證明新的 appearance module 在 Windows 上編得過也沒留 warning，這塊在 mac 上驗不到。

**我動了你的分支。** 主線合併 #312 之後 `CHANGELOG-NEXT.md` 撞到了，我用 maintainer 權限把 master 合進來、手動解掉，順便在貢獻者段補上你的名字。用的是 merge 不是 rebase，所以你的 commit 沒被改寫，直接 pull 就好，不用擔心歷史對不上。

回頭看上一輪那份 review，該處理的你幾乎全中了。幾個我特別想提的：

預覽那條你不是補個按鈕就算了，是整段改成暫存的 draft store，選檔案只記路徑、預覽直接讀來源檔、按下 Apply 才真的寫進 app data。這個做法比我原本要求的「加一個 pre-commit 預覽步驟」乾淨得多。

拖放那塊你做得比我建議的好。我只說了「共享一個 registry」，你連 priority 一起做進去，順手把 `TerminalView` 那個 `document.querySelector` 的跨模組耦合跟重複的訂閱邏輯都收掉了。

DPI 你沒照我說的「一律除以 devicePixelRatio」，改成平台分支，這個判斷是對的，因為你在 mac 上實測過 Retina 的實際行為，比我照文件推論可靠。不過 Windows 那一半目前還是推論，沒有人實際在 Windows 上拖過圖，CI 也驗不到這種行為。失敗模式很溫和（最多圖沒進去），所以我先合了，有人回報再修。

Rust 那邊 `sync_all`、`.tmp` 排除、`read_to_end`、刪檔重試全部到位，本機跑 `modules::appearance` 9 個測試也都過。

還有一個小洞我合併後自己補，不用你動：`AboutSettingsSection.tsx:123` 也會呼叫 `setSettingsOpen(false)`，但沒有清 draft。`SettingsModal` 的 close、Escape、preview 面板那顆按鈕都有 `cancel()`，就這條漏掉。後果不嚴重，那時 `previewActive` 是 false 所以背景不會錯，只是下次打開設定會看到上次沒套用的變更還在。在 `SettingsModal` 加個 unmount cleanup 就能把所有繞過路徑一次蓋掉。

PR #327, opened 2026-08-12
feat(editor): VS Code-style scrollbars for the editor and diff views

yw-chan:
## Problem

On Windows, WebView2 only has always-on classic scrollbars, which made scrolling in the code surfaces awkward in three ways:

1. The horizontal scrollbar spans the whole scroller, running beneath the sticky line-number gutter — and since the gutter background was transparent, long lines and diff line tints also slid visibly under the numbers while scrolling.
2. In the diff tab the merge editors are auto-height (the outer `.cm-mergeView` owns vertical scrolling), so each side's horizontal scrollbar sat at the bottom of the whole document — visible only after scrolling all the way down.
3. The two diff sides scrolled horizontally independently.

## Root cause

- The syntax themes deliberately keep the editor background transparent so the app background shows through, but the gutter is `position: sticky` inside the horizontal scroller, so a transparent gutter lets content slide through beneath it.
- While fixing that we found CodeMirror mounts theme style modules with *earlier-extension* precedence (the `styleModule` facet is reversed before mounting), so the existing one-dark transparent-background overlay and unified active-line color, placed *after* `oneDark`, never actually took effect.

## Changes

- **themes**: paint the gutter with the app background token (`var(--color-bg)`) — visually identical to transparent, but opaque, so nothing shows through beneath the line numbers. Move the one-dark overrides before `oneDark` so they win the cascade.
- **`src/lib/proxyScrollbar.ts`** (new, Windows-only via `IS_WINDOWS`; macOS/Linux keep the stock native overlay scrollbars): hide the scroller's own bars and replace them with thin overflow strips the browser renders real native scrollbars for, placed like VS Code — horizontal pinned to the pane bottom starting after the gutter (or right under the content when a short diff ends inside the pane), vertical at the editor's right edge. Strips and scroller get equal scroll ranges so positions map 1:1 with no math on scroll events, and the horizontal strip hides entirely when nothing overflows (e.g. word wrap on). The host class lives on the editor's wrapper rather than `view.dom`, which CodeMirror rewrites on update.
- **diff**: sync horizontal scrolling between the two sides, 1:1 like VS Code — the side with shorter lines stops at its own end while the other keeps going (applies on all platforms).

## Test plan

- [x] `npx pnpm exec tsc --noEmit`
- [x] `npx pnpm exec vitest run` — 1988 passed
[...]

PR #348, opened 2026-08-25
fix(editor): paint the wallpaper into the gutter instead of leaking code through it

yw-chan:
## Problem

With a custom background image set, horizontally scrolling the editor or the diff view drags code straight through the line-number gutter — numbers and code render on top of each other. Both CodeMirror surfaces are affected, and in the diff view both sides show it.

## Root cause

#327 gave the sticky gutter an opaque backdrop precisely so scrolled code stops sliding through beneath the numbers. #342 then found that backdrop compounding over a wallpaper — the editor's wrapper already paints one `--color-bg` layer, a second one on the gutter composites to ~97% against the wrapper's ~82% — and made the gutter transparent under `.wallpaper-surface`. That fixed the stripe and, as its message noted, let the pre-#327 bleed back in for wallpaper mode. This is that bleed.

Two things that do **not** work, for the record:

- Painting a flat backdrop back on, even only while scrolled: that is #342's stripe again.
- Clipping the scrolled-past content from JS. Scrolling runs on the compositor, so by the time a scroll event lands the frame is already on screen and the clip edge trails it by one; dragging the proxy scrollbar and syncing the diff's other side add another frame each. It visibly jitters along the gutter edge.

## Changes

The gutter paints the wallpaper itself, plus the same tint the wrapper uses. It is opaque, so the code beneath it stays hidden; it is the same image and the same tint as the surface beside it, so there is no stripe; and it is completely static — nothing tracks the scroll position, so there is nothing to lag.

- `themes/editorTheme.ts`: `.cm-gutters` gains a `--cm-gutter-bg-image` layer pinned with `background-attachment: fixed`, which positions from the viewport rather than from the gutter's own scrolled box.
- `index.css`: `.wallpaper-surface` defines that token as `linear-gradient(tint, tint), var(--wallpaper-image)`. Outside wallpaper mode the token is undefined, the declaration drops out, and the gutter keeps the opaque colour #342 left it with.
- `BackgroundImageLayer.tsx`: `fixed` can only be told about the viewport, and a workspace-scoped wallpaper covers just the workspace column — so the layer publishes where `object-fit: cover` actually put its image, in viewport coordinates, as `--wallpaper-fixed-size` / `--wallpaper-fixed-pos`. Measured on layout changes only (a `ResizeObserver` plus the image's own `load`), never on scroll.

Both surfaces pick this up through `editorSyntaxTheme`, so the editor and the diff view are covered.

## Test plan

- [x] `npx pnpm exec vitest run` — 2026 passed
- [x] New tests: the gutter's wallpaper layer is pinned to the viewport and adds no colour of its own; the published geometry matches what `object-fit: cover` does
- [x] Manual (Windows, background image set): scroll the editor and both diff sides horizontally — with the mouse wheel and by dragging the bottom proxy scrollbar — and confirm no code shows through the numbers, no darker stripe appears down the gutter, and nothing jitters. Repeat with no background image to confirm the gutter is unchanged.

### DIFF
--- a/src/components/BackgroundImageLayer.test.tsx
+++ b/src/components/BackgroundImageLayer.test.tsx
@@ -34,6 +34,38 @@ describe("BackgroundImageLayer", () => {
     expect(screen.queryByTestId("background-image-window")).toBeNull();
   });
 
+  it("publishes where the image landed so the gutter can pin the same copy", () => {
+    useSettingsStore.setState({
+      backgroundImagePath: "/app-data/appearance/background.png",
+      backgroundImageOpacity: 20,
+      backgroundImageScope: "workspace",
+    });
+    const { unmount } = render(<BackgroundImageLayer scope="workspace" />);
+    const image = screen.getByTestId("background-image-workspace");
+    Object.defineProperty(image, "naturalWidth", { value: 200, configurable: true });
+    Object.defineProperty(image, "naturalHeight", { value: 100, configurable: true });
+    image.getBoundingClientRect = () =>
+      ({ left: 40, top: 20, width: 400, height: 400 }) as DOMRect;
+
+    act(() => {
+      image.dispatchEvent(new Event("load"));
+    });
+
+    // cover on a 400x400 box scales the 200x100 source by 4, so it overflows
+    // 200px to each side of a box that itself starts 40px into the viewport.
+    const root = document.documentElement.style;
+    expect(root.getPropertyValue("--wallpaper-fixed-size")).toBe("800px 400px");
+    expect(root.getPropertyValue("--wallpaper-fixed-pos")).toBe("-160px 20px");
+    expect(root.getPropertyValue("--wallpaper-image")).toBe(
+      'url("/app-data/appearance/background.png")',
+    );
+
+    unmount();
+    expect(root.getPropertyValue("--wallpaper-image")).toBe("");
+    expect(root.getPropertyValue("--wallpaper-fixed-size")).toBe("");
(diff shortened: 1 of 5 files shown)


mukiwu:
深審過了，合。幾何算式用具體數字對過（400×400 框、200×100 圖 → 800×400 @ -160,20），`.cm-gutters` 目前的祖先鏈沒有 transform / filter / will-change，`background-attachment: fixed` 對得上；非桌布模式下 token 未定義、宣告整條被丟掉、行號區維持 #342 留下的不透明色，也確認過。

三個非阻擋的加固建議，不影響這次合併，之後順手或另開小 PR 都可以：
1. `editorTheme.ts` 的 gutterWallpaper 規則加一行註解，列出祖先不能有的屬性（`transform`、`filter`、`will-change`、`perspective`、`contain: paint`），不然日後有人加了對齊會悄悄壞掉。`SessionsPanel.tsx:107` 有類似先例
2. `BackgroundImageLayer.tsx:42` 的 `url("${src}")` 對 `"` 與 `\` 做一下跳脫。現在 `convertFileSrc` 出來的路徑不會有這些字元，純防禦
3. `background-attachment: fixed` 在 WebView2 上捲動會逐幀重繪，行號區很小應該沒感覺，方便的話在 Windows 開個一萬行的檔橫向捲一下確認

另外 review 時有懷疑過 window / workspace 兩個圖層切換範圍時會不會互清變數，看過之後不成立：兩層訂閱同一個 store、同一個 commit 更新，React 在一個 commit 內一律先跑完 cleanup 再跑新的 effect，順序固定是舊清新寫。

PR #350, opened 2026-08-25
fix(tabs): make an overflowing tab strip usable

yw-chan:
## Problem

Open enough tabs and the strip overflows. On Windows the browser answers that with a classic horizontal scrollbar, which takes real height out of a 36px row: every tab is squashed to make room and the bar itself sits somewhere it clearly does not belong.

Behind that, everything else about an overflowing strip was untested ground, and most of it was broken:

- The wheel did not scroll it.
- The add-tab button scrolled away with the tabs, out of reach.
- Middle-click-to-close stopped working.
- Activating a tab past the edge left it off-screen.

## Root cause

Each has its own cause, and they only appear once the strip actually scrolls:

- **The wheel.** A browser only turns a vertical wheel sideways once the box can scroll nothing else. This one can: the tabs stretch it to the row's full height.
- **Middle-click.** A scrolling strip is a scroll container, and Chromium answers a middle-click on one with autoscroll instead of delivering the `auxclick` the close gesture listens for.
- **The add-tab button.** It lived inside the scrolling strip, after the last tab.
- **The scrollbar itself was undraggable** even once styled: `data-tauri-drag-region` was on the strip, and Tauri swallows mousedown on a drag region to start moving the window, so the thumb never got the press.

## Changes

- **A 3px hairline scrollbar** (`index.css`): transparent track, `--color-border-strong` thumb, brighter on hover. Only visible when the tabs actually overflow. This is the file's one `::-webkit-scrollbar` rule, against the note at the top of that section — and it is keyed off `data-tab-strip="hairline"`, which `TabBar` sets only off macOS. WKWebView's overlay scrollbars are already thin, transient and free of layout cost; a pseudo-element rule is exactly what would trade them for an always-present bar that takes height, i.e. inflict this bug on the one platform that never had it. Windows and Linux, which draw classic bars, get the hairline.
- **A plain wheel scrolls the strip sideways; Shift+wheel steps through the tabs**, one per mouse notch, stopping at either end rather than wrapping — a wheel has no detent to tell you that you just crossed from the last tab back to the first. A trackpad's finer deltas accumulate to the same step, and its own sideways swipe (`deltaX`) is left to the browser, inertia and all. Registered by hand because React marks wheel listeners passive and both branches cancel the default.
- **The strip now takes only the width its tabs need**, with the add-tab button after it and the leftover slack after that. The button keeps its place right behind the last tab, and overflowing tabs scroll *past* it rather than over it. Pinning it inside the strip works too, but only with a backdrop for the tabs to slide under — and over a background image that second layer of the row's own tint reads as a lighter patch, the compounding #342 fixed for the editor gutter. `data-tab-bar` moves out to a wrapper around all three, so the drop target the drag stores resolve against covers the same ground as before; the strip carries `data-tab-strip`.
- **`data-tauri-drag-region` comes off the strip**, so its scrollbar can be dragged. The slack beside it carries the drag region instead — the part of the row that was ever worth grabbing.
- **Activating a tab past either edge brings it in**, by the shortest distance that makes it whole. Positioned by hand from rects rather than with `scrollIntoView`: WKWebView's is unreliable inside nested scroll containers (the same reason `NoteToc` positions its container itself), and it would be free to scroll ancestors, which here means yanking the whole shell sideways.
- **Middle-click's autoscroll is cancelled** on the strip, covering its tabs and its empty space alike. `auxclick` fires on mouseup, so the close still lands.

macOS gets the wheel behaviour and the layout; it keeps its native overlay scrollbar, never had middle-click autoscroll (that `preventDefault` is a no-op there), and is specifically why the scroll positioning avoids `scrollIntoView`.

## Test plan

- [x] `npx pnpm exec vitest run` — 2033 passed
- [x] New tests: hairline requested off macOS and withheld on it; drag region off the strip; wheel scrolls; shift+wheel steps and stops at the ends; a sub-step delta accumulates instead of stepping; add button outside the strip but inside the drop target; activating an off-screen tab scrolls it in by the exact distance, and an on-screen one not at all
- [x] Manual (Windows): open enough tabs to overflow, then check the tabs keep their full height, the hairline appears and drags, the wheel scrolls, Shift+wheel walks the tabs and stops at both ends, middle-click still closes, the add button stays reachable, and Ctrl+Tab pulls an off-screen tab into view
- [ ] Manual (macOS): same, minus the hairline — the overlay scrollbar should be untouched

### DIFF
--- a/src/components/TabBar.test.tsx
+++ b/src/components/TabBar.test.tsx
@@ -11,6 +11,16 @@ import { useSshDragStore } from "@/modules/ssh/lib/sshDrag";
 import { useUiStore } from "@/stores/uiStore";
 import { useWorkspaceStore } from "@/stores/workspaceStore";
 
+// IS_MAC is a module-load const; expose it through a getter so a test can flip
+// the platform without re-importing the module.
+const platformMock = vi.hoisted(() => ({ isMac: false }));
+vi.mock("@/lib/platform", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@/lib/platform")>()),
+  get IS_MAC() {
+    return platformMock.isMac;
+  },
+}));
+
 beforeEach(() => {
   useTabsStore.setState({
     spaces: [{ id: "s1", name: "One" }],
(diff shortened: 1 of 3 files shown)


mukiwu:
深審過了，乾淨，直接合。滾輪方向與 deltaMode、Shift+滾輪累積器在 0 與反向時的行為、中鍵 `preventDefault` 與 dnd-kit 的先後順序（pointerdown 先於 mousedown）、四個拖放 store 的 `closest("[data-tab-bar]")` 在屬性搬到外層後涵蓋範圍不變、溢位時 + 按鈕的 flex 佈局、macOS 紅綠燈 `pl-20` 拖曳區、hairline 只在非 macOS 套用，都逐一驗過。

幾個非阻擋的小備註，不用改，記著就好：
- 兩個滾輪分支的 `preventDefault` 不管有沒有真的捲到都會呼叫。這個列上面沒有可捲動的祖先所以無害，哪天搬進可捲動容器再加判斷
- Shift+滾輪的 `pending` 只在反向時歸零，停頓很久再轉可能會多跳半格；Windows 上用起來覺得怪再加時間重置
- 切分頁那個測試在 jsdom 只驗得到算式，真實 clamp 靠你的 Windows 手測

謝謝這個 PR，問題拆解跟根因寫得很清楚。

PR #357, opened 2026-08-31
feat(git-graph): merge a branch with its remotes into one ref chip

yw-chan:
## Problem

A commit that has `master`, `origin/master` and `origin/HEAD` on it draws three
chips of near-identical information. Add a couple of other branches and a row
easily carries four or five, and the chip strip runs longer than the commit
message — which gets pushed out of view entirely.

`origin/HEAD` is the worst offender: it is a symbolic link to the remote's
default branch, so it marks a commit that already has its own branch chip, and
it costs a permanent chip slot to say nothing.

The right-click menu is split along the same lines. Local operations (checkout,
merge, delete) live on the `master` chip, while remote ones (pull, delete the
remote branch) live on `origin/master`. The user has to know which chip owns
which half before they can right-click.

## Changes

- `lib/refChips.ts` — a pure `buildRefChips(refs, options)` that folds a local
  branch and the same-named remotes at that commit into one chip, drops
  `origin/HEAD`, and splits off whatever exceeds the row's limit. Git does not
  promise the local ref precedes its remote twin in the decoration, so remotes
  seen first are held back and the merged chip keeps the local ref's slot. A
  detached `HEAD` is never folded into `origin/HEAD`.
- `RefChipStrip.tsx` — a merged chip is drawn as blocks inside one border: the
  branch, then each remote in italic behind a divider. Everything past the limit
  collapses into a `+N` chip that opens a portal list of the same chips, each
  still right-clickable; the list sits below the tooltip layer so the chips'
  tooltips stay visible over it, and it stays open behind the menu it opens.
  Only the ref kinds that actually have a menu get a hover state.
- `lib/contextMenuItems.ts` — `buildRefMenu` takes the merged remotes, so one
  menu runs checkout / merge / worktree, then pull per remote, then both
  deletions, then copy. The current branch (`head`), which had no menu at all,
  gets its remotes' actions once they are folded in. `onPull` / `onDeleteRemote`
  now take a remote ref name, since a merged chip can carry several.
- A remote with no local twin at that commit — the ahead/behind case — is
  untouched: it keeps its own `origin/…` chip on the commit it really points at,
  so nothing about the divergence is lost.
- Settings > Git Graph (new section): merging, hiding `origin/HEAD` and the `+N`
  threshold are each switchable, all on by default with a limit of 3. Turning
  all three off restores exactly the previous row, so the change breaks nobody's
  habits. The persisted block is normalized on hydration, so a store written by
  an older build gets the defaults and a stray limit is clamped to 1–10.

[...]

PR #360, opened 2026-09-01
feat(git-graph): copy a ref name from every branch and tag chip

### LINKED ISSUE #332: [許願] git graph 右鍵選單新增 `複製 branch name`
git graph 右鍵選單新增 `複製 branch name`

Image

yw-chan:
## Problem

`Copy branch name` was only reachable from a remote chip, or from a local chip
that had a remote folded into it. A branch that has not been pushed yet — the
one whose name you most often want on the clipboard, to paste into a checkout
or a PR — had no way to copy it.

The chip for the branch you are standing on was worse: with no remote merged in
it produced an empty item list, and the graph skips an empty menu, so
right-clicking the current branch did nothing at all.

A tag chip has the same gap in a smaller way: its only entry is the red
`Delete tag`, so a right-click on `v1.0.0` offers a destructive action and
nothing else.

## Root cause

`buildRefMenu` gated the copy entry on `remotes.length > 0`, which happens to be
true for every chip the author was looking at while building the merged chip in
#357 — a local branch with a remote twin. The `head` path returned early with an
empty list whenever nothing was folded in, on the reasoning that checkout, merge
and delete do not apply to the branch already checked out. Copy does.

## Changes

- `lib/contextMenuItems.ts` — copy is now unconditional on a branch chip, so a
  local-only branch and the current branch both have it. The early return is
  replaced by a narrower guard: a detached `HEAD` still has no menu, because
  nothing here acts on it and its name is the literal `HEAD`.
- A tag chip gains `Copy tag name`, in its own group so a divider fences it off
  from the delete above it.
- `onCopyBranchName` is renamed `onCopyRefName`: one handler now serves the
  branch, the remote ref and the tag, and it has always just copied `ref.name`.
- New `menu.copyTagName` string in `en` and `zh-Hant`.

Closes #332.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm exec vitest run` — 227 files, 2080 tests passing (`contextMenuItems`
      covers the local-only branch, the current branch, the detached `HEAD` and
      the tag chip)
- [x] `tauri dev` on Windows: right-click an unpushed branch chip, the green
      current-branch chip and a tag chip, and confirm each copies its own name

### DIFF
--- a/src/i18n/locales/en/gitGraph.json
+++ b/src/i18n/locales/en/gitGraph.json
@@ -74,6 +74,7 @@
     "pullFrom": "Pull from {{remote}} (git pull)",
     "deleteRemoteOn": "Delete branch on {{remote}}… (git push --delete)",
     "copyBranchName": "Copy branch name",
+    "copyTagName": "Copy tag name",
     "openWorktree": "Open worktree for this branch"
   },
   "modal": {
--- a/src/i18n/locales/zh-Hant/gitGraph.json
+++ b/src/i18n/locales/zh-Hant/gitGraph.json
@@ -74,6 +74,7 @@
     "pullFrom": "從 {{remote}} 拉取 (git pull)",
     "deleteRemoteOn": "刪除 {{remote}} 上的分支… (git push --delete)",
     "copyBranchName": "複製分支名稱",
+    "copyTagName": "複製標籤名稱",
     "openWorktree": "為這個分支開 worktree"
   },
   "modal": {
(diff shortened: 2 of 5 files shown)


mukiwu:
深審過了，乾淨，合。rename 全庫無遺漏、detached HEAD 的判斷跟 refChips 的既有慣例一致、順帶把 #357 那條「目前分支右鍵空選單」修掉了，測試也都是行為層。謝謝。

PR #361, opened 2026-09-01
feat(git-graph): order a commit's ref chips by what the row is read for

yw-chan:
## Problem

The chips on a commit row appear in the order git hands them over, and git
lists decorations in the reverse of the order it walked the refs. In practice
that puts `refs/remotes/…` ahead of `refs/heads/…`:

```
(refs/remotes/origin/fix/tab-bar-overflow-scroll, refs/heads/fix/tab-bar-overflow-scroll)
```

Reverse ref-name order is an implementation detail with nothing to say to a
reader, and it was deciding two things it should not. It picks what the eye
hits first on the row, and — since #357 — it picks which chips survive
`collapseAfter` and which fold into `+N`, because that is a plain
`slice(0, limit)` over git's order. A row can therefore hide a local branch
behind a remote chip that carries no information the branch does not.

## Changes

- `lib/refChips.ts` — `compareChips` ranks by ref kind: `head`, then `branch`,
  then `tag`, then `remote`, then `stash` / `unknown`. That is the order the row
  is read in — where HEAD is, then the branches you act on, then the milestones,
  then the remotes with no local twin at this commit (the ahead/behind case),
  then the read-only decorations. Ties fall back to the ref name, so a row keeps
  the same shape across refreshes.
- `origin` leads its own kind, since it is the remote nearly every repo pushes
  to. The same rule sorts the blocks inside a merged chip, so it reads
  `master │ origin │ upstream` rather than in whatever order git listed them.
- The sort runs after the folding, so a merged chip is ranked as the branch it
  is rather than as whichever half of it git happened to list first.
- No new switch. The three Git Graph options govern what a row shows; the order
  it shows it in is not a preference anyone holds. This does narrow #357's
  "turn all three off and you get exactly the previous row" slightly: with the
  options off you still get every ref, unmerged, `origin/HEAD` included — just
  not in git's reverse-alphabetical order. `refChips.test.ts` states that
  explicitly now.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm exec vitest run` — 227 files, 2083 tests passing (4 new in
      `refChips`: the kind ranking, name and `origin`-first ordering within a
      kind, `origin`'s block leading a merged chip, and `+N` collapsing the
      least important refs rather than whichever git listed last)
- [x] Two existing tests were about git's order and are rewritten, not deleted:
      the merged-remote case now asserts that a remote listed before its local
      twin gets folded rather than given a chip of its own, and the
      options-off case asserts the refs are all still there. The `RefChipStrip`
      overflow fixtures are renamed `a`–`e` so they stop depending on the order
      by accident.
- [x] `tauri dev` on Windows: a commit carrying HEAD, a branch, a tag and a
      lone remote renders in that order, and a row past the threshold keeps the
      branches on the row

### DIFF
--- a/src/modules/git-graph/RefChipStrip.test.tsx
+++ b/src/modules/git-graph/RefChipStrip.test.tsx
@@ -152,38 +152,38 @@ describe("RefChipStrip", () => {
     render(
       <RefChipStrip
         refs={[
-          ref("branch", "one"),
-          ref("branch", "two"),
-          ref("branch", "three"),
-          ref("branch", "four"),
-          ref("branch", "five"),
+          ref("branch", "a"),
+          ref("branch", "b"),
+          ref("branch", "c"),
+          ref("branch", "d"),
+          ref("branch", "e"),
         ]}
         options={OPTIONS}
         labels={LABELS}
       />,
     );
     fireEvent.click(screen.getByText("+2"));
-    const list = screen.getByText("five").closest("div") as HTMLElement;
+    const list = screen.getByText("e").closest("div") as HTMLElement;
 
     // The list itself is max-h + overflow-y-auto: reading past its fold is a
     // scroll event whose target is the list. Capture-phase listeners on window
     // see it too, and closing on it would make the tail unreachable.
     fireEvent.scroll(list);
-    expect(screen.queryByText("five")).not.toBeNull();
+    expect(screen.queryByText("e")).not.toBeNull();
 
     // The graph scrolling underneath is what the dismissal is for.
     fireEvent.scroll(document);
-    expect(screen.queryByText("five")).toBeNull();
+    expect(screen.queryByText("e")).toBeNull();
   });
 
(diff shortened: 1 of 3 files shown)


mukiwu:
深審過了，排序的每個宣稱都驗過：sort 在摺疊後、slice 前，所以 +N 收的是最不重要的；合併 chip 以本地那半排名；origin 的特例只作用在 remote 對 remote；選單裡每個 remote 仍個別可點，重排不會改變任何動作的對象。測試拿掉改動會紅。

一個小東西我直接在分支上補了一個 commit：名字平手時原本用無參數的 `localeCompare`，它跟著執行環境的預設 collator 走，Node（跑測試的）跟 WKWebView（使用者看到的）理論上可能排出不同順序。三處都換成單純的 codepoint 比較（`byCodepoint`），到處都是同一個結果；分支名幾乎都是 ASCII，沒有語言學排序的損失。CI 綠了我就合。

PR #364, opened 2026-09-01
feat(source-control): reveal row actions on hover, pin the file being viewed

yw-chan:
## Problem

The Source Control panel is narrow and the paths in it are long, yet every file row pinned its stage/unstage/discard buttons at the trailing edge. Names truncated against a permanent gutter of buttons even with the pointer nowhere near the row.

Two gaps came with it: nothing marked which file's diff was actually on screen, and the Staged changes header had no counterpart to the "Stage all" action on Changes.

## Changes

- **Row actions are revealed on hover.** File rows, folder headers and section headers keep their actions in a strip collapsed to zero width until the row is hovered — the same pattern `SessionsPanel` already uses. Width (not `display`) keeps the buttons in the tab order, `group-focus-within` expands the strip when one is tabbed to, and clipping lifts on expand so the global `:focus-visible` ring is not cut off. Every action hidden this way is also in the row's context menu.
- **The file being viewed stays pinned.** The row whose diff sits in the foreground pane keeps its highlight and its actions out, and reports `aria-current`. A new `activeDiffPane()` in `tabsStore` reads the active leaf of the active tab, mirroring the existing `activeEditorPath` rule, so a split reports the focused pane. `staged` is half of the match: a part-staged file appears in both lists, and only the side actually on screen is the current one. It is read as two primitive selectors — one returning a fresh `{path, staged}` would never compare equal and would re-render the panel on every store change.
- **Unstage all.** The Staged changes header gains a `Minus` button that unstages every staged file, the counterpart of "Stage all"; `unstageAll` added to the `en` and `zh-Hant` locales.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm test` — 227 files, 2080 tests passing
- [x] Three new tests in `SourceControlView.test.tsx`: the foreground diff's row is marked current, only the on-screen side of a part-staged file is marked, and the staged header unstages every staged file
- [x] Manual pass in the running app (hover reveal, the highlight following the foreground diff) — not run on my machine yet; the change is CSS plus the new store selector, both covered above

### DIFF
--- a/src/i18n/locales/en/sourceControl.json
+++ b/src/i18n/locales/en/sourceControl.json
@@ -6,6 +6,7 @@
   "changes": "Changes",
   "noChanges": "No changes",
   "stageAll": "Stage all",
+  "unstageAll": "Unstage all",
   "stage": "Stage",
   "unstage": "Unstage",
   "viewFolder": "Group by folder",
--- a/src/i18n/locales/zh-Hant/sourceControl.json
+++ b/src/i18n/locales/zh-Hant/sourceControl.json
@@ -6,6 +6,7 @@
   "changes": "變更",
   "noChanges": "沒有變更",
   "stageAll": "全部暫存",
+  "unstageAll": "取消全部暫存",
   "stage": "暫存",
   "unstage": "取消暫存",
   "viewFolder": "依資料夾分組",
--- a/src/modules/source-control/SourceControlView.test.tsx
+++ b/src/modules/source-control/SourceControlView.test.tsx
@@ -52,6 +52,34 @@ describe("SourceControlView row interactions", () => {
     expect(tabs[0].title).toBe("a.ts");
   });
 
+  it("marks the row whose diff is in the foreground as the current one", async () => {
+    render(<SourceControlView />);
+    const row = await screen.findByText("src/a.ts");
+    expect(row.closest("li")).not.toHaveAttribute("aria-current");
+
+    fireEvent.click(row);
+
+    expect(screen.getByText("src/a.ts").closest("li")).toHaveAttribute("aria-current", "true");
+  });
+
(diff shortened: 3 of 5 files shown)


mukiwu:
深審過了，核心都乾淨：RowActions 忠實復刻 SessionsPanel 的模式、`activeDiffPane` 的窄化與兩個 primitive selector 正確、不需要動 persist 版號、i18n 兩個 locale 齊、part-staged 的判斷本來就對稱（我補了 staged 側的測試，直接綠）。

兩件事我直接在分支上修了（一個 commit）：

1. **hover 展開只留給檔案列，區段與資料夾標題的動作改回常駐**。「藏起來的動作右鍵選單都找得到」只對檔案列成立，資料夾列跟區段標題沒有 context menu，包進 hover 條之後，平板模式或點按式觸控板會完全按不到 Stage all／Unstage all／資料夾整批操作；而且 Stage all 原本一直是常駐的，肌肉記憶會撲空。標題每列只有一顆按鈕，常駐不佔什麼寬度，檔案列的收合（這個 PR 的重點）原封不動。
2. **Unstage all 改成跟 Stage all 同款的文字按鈕**。原本一邊是強調色文字、一邊是灰色小圖示，看起來不像對等物。

測試補了兩條：標題動作不收合、檔案列照樣收合（class 層面的契約，jsdom 不套 Tailwind 所以測不了視覺）、part-staged 開 staged 側 diff 時標記在 staged 列上。全套 2082 項綠、typecheck 過。

殘留的已知空白跟你 test plan 寫的一樣：hover 展開的實際視覺行為（jsdom 測不到）你我都還沒在跑起來的 app 裡確認過，不過這個模式 SessionsPanel 已經出貨用了一陣子，而且這次改動後常駐的東西變多、hover 相依變少，風險是往下走的。CI 綠了我就合。

PR #406, opened 2026-09-05
feat(diff): read every uncommitted change on one page

yw-chan:
Closes part of #397. This is the first of three; see **Scope** below.

## Problem

Reading through what you have not committed yet means one tab per file. Eight
files, eight tabs, and the tab strip runs out of room before the review does —
so the only way to see the next file is to go back to the panel, click again,
and open another tab. There is no view that answers "what have I changed?" in
one place.

## Changes

A new **All Changes** tab: every uncommitted file's diff stacked on one page,
staged and unstaged as two sections, read by scrolling.

- **Scan.** One `gitStatus` plus two `gitDiff` calls give the file list and the
  per-file `+`/`−` counts up front, so every section header is filled in before
  any editor exists. `parseDiffStats` tracks whether it is inside a hunk, so a
  diff of a diff does not derail the count.
- **Lazy mount.** An IntersectionObserver with an 800px margin builds the
  CodeMirror views for the sections near the viewport and tears down the ones
  that leave it. A 77-file repo mounts three editors, not 77.
- **Truncation.** A file with more than 500 changed lines starts folded, header
  and counts only — a lock file should not eat the page. Its header opens it.
- **Toolbar.** Previous/next change, word wrap, comment and send-to-agent are
  the same four controls the single-file diff tab has, reused rather than
  rebuilt. Previous/next walks changes across files, from a hunk index built
  out of the same scan.
- **Entry points.** A toolbar button in the Source Control panel, View → All
  Changes, and ⇧⌘G / Ctrl+Shift+G.

The single-file diff tab is refactored, not changed: the merge-view builders,
the comment loop and the send-to-agent action move to `src/modules/diff/lib/`
so both surfaces share them. `DiffTabContent.tsx` loses 370 lines and gains
nothing.

## On the three loose ends in #397

**Shutting a file must not cost a diff — already the case.** A file the scan
says is oversized never loads its two documents, so no `MergeView` is built and
no diff is computed until the reader opens it
(`DiffFileSection.tsx:142`), and its placeholder height is estimated from the
hunks instead (`:322`). The one file that still loads is an untracked one,
because `git diff` does not report it and there is no other way to know its
size — and that case is safe: its index side is empty, so the comparison
degenerates to a single insertion and never approaches the timeout.

**The no-op `diffConfig` is gone.** `reconfigure()` reads it as
`"diffConfig" in config`, so leaving the key out keeps the one the view was
built with; the line is removed with half a comment saying why, in
`lib/diffViews.ts`.

**#401's comment is accurate here.** It describes a file scrolling out of the
mounted window, which is this view — the thing it pointed at now exists.

Still outstanding: the inline path has no regression test of its own for the
diff config. Happy to add it here or send it separately, whichever you prefer.

## Scope

Two things go beyond what the issue thread settled, both additive:

- **Per-file collapse** (`feat(diff): open and shut each file from its own
  header`) — the chevron is in the issue's own figure, but the thread only
  agreed the size-based rule. Say the word and it comes out.
- **Page ordering** follows the panel's tree (folders before files,
  alphabetical) rather than git's output order. Internal to this new view;
  nothing existing changes.

One change touches behaviour outside this page: `fix(diff): hold the fold
gutter's width` gives `.cm-diff-fold-gutter` a fixed 18px. #393 deliberately
left it unsized, but a gutter is only asked for markers over the *visible*
range — scroll a fold icon into view and every line of code shifts right,
scroll it away and they shift back. Holding the width costs 18px on a file
that folds nothing and keeps the text still on every file that does. This
merged cleanly with #404 and is the one commit worth a second opinion.

## Test plan

- `npx tsc --noEmit` — clean.
- `npx vitest run` — 2,153 pass across 235 files, nothing skipped.
- New coverage: 12 for the page, 8 for `parseDiffStats` (a diff-of-a-diff among
  them), 4 for the binary search behind the change counter. Each of the three
  scroll fixes was proved by disabling the fix and watching its test go red.
- By hand on a 77-file working tree: scrolled top to bottom, watched only the
  sections near the viewport hold editors; folded and unfolded a 3,000-line
  lock file; walked previous/next across file boundaries; sent comments from
  two different files in one batch; shut a file from half way down it; checked
  the header at a 400px pane.

Rebased onto master after #400, #401, #402 and #404. #402's fix moved with the
code it was in: it now lives in `lib/diffViews.ts` and applies to the
all-changes editors too, and its test on `DiffTabContent` is untouched.


mukiwu (approved):
已完成最終審查：All Changes 的大範圍 diff、lazy mount、跨檔案導覽與既有 comment / send-to-agent 流程符合已確認的功能方向；ignored status 已由 #413 先修正，quoted / non-ASCII Git path 也已補解析與測試。這是中大型 additive feature，但沒有新增依賴或改寫 TempoTerm 核心 pane / workspace 模型，前端與 Windows CI 均通過。

PR #407, opened 2026-09-05
fix(git-graph): make the remote, tag and stash toggles actually do something

yw-chan:
## Problem

The Git Graph toolbar has three display toggles — Show Remote Branches, Show Tags, Show Stashes — and none of them worked.

- **Show Remote Branches / Show Tags**: turning either off looked completely inert. Every `origin/x` and `tag: v1` chip stayed on the row. When a remote branch *was* ahead of local, it was worse than inert: those commits vanished from the graph while the chip stayed behind, painted on an older commit.
- **Show Stashes**: turning it on did nothing at all, in any repository.

## Root cause

All three toggles only ever fed `git log`'s rev range, and a rev range is not what a reader sees.

`--decorate` paints a ref on any commit the walk already reached, no matter how it got there. Remote branches and tags nearly always sit on commits a local branch can reach, so dropping `--remotes` / `--tags` from the range removes almost nothing while every chip survives.

`--glob=refs/stash` never matched anything. Git turns a glob with no wildcard into a prefix, so it expanded to `refs/stash/*`. Adding the missing `*` would not have been a fix either: `refs/stash` only names `stash@{0}` — the older entries live in that ref's reflog, not in refs of their own — and walking from it drags git's internal `index on ...` and `untracked files on ...` snapshots into the graph as if the user had committed them.

## Changes

- Filter ref decorations by the same toggle that filters the range, so a chip can never outlive the commits its toggle removed.
- Read stashes from `refs/stash`'s reflog (`git rev-list --no-walk --parents -g refs/stash`) and list them by hash, so every entry appears and not just the newest. Drop the index/untracked helper commits that walk pulls in, and compensate `--max-count` for them so paging still reports `has_more` correctly.
- Label each stash `stash@{n}`, matching how `git stash list` names them. Git only decorates `refs/stash`, so without this the older entries render as a row of unlabelled "WIP on ..." commits.
- Drop the helper commits from their stash's parents as well. A parent pointing at a commit that is not in the graph is dangling, and the lane layout reserves a track for every parent and then waits forever for one that never arrives. Four stashes were enough to push the lanes past `maxLane`, where they collapse onto a single column: two nodes landed on the same x and the graph drew a parent/child line between two commits that are siblings.
- `GraphOptions::default()` derived all-false, which now means "decorate nothing". Give it the toolbar's own defaults instead — remotes and tags on, stashes off.

Behaviour deliberately left alone: when specific branches are picked in the branch filter, the range still ignores the three toggles, so a filtered graph does not union in every remote branch's history. Decoration filtering applies in both modes.

## Test plan

- [x] `cargo test --lib` — 521 passed. Six failures are pre-existing on `master` on this Windows box (one `core.autocrlf` assertion, five pty tests spawning `/bin/cat`); verified by re-running them against `master`'s copy of the file.
- [x] `pnpm typecheck` — clean.
- [x] `pnpm test` — 2127 passed (232 files).
- [x] New unit tests: reflog parsing splits stash tips from helper parents, decoration filtering per toggle, `stash@{n}` labelling, and the toolbar-matching defaults.
- [x] New integration tests against real repositories: every stash shows with no helper commits and no dangling parents; remote and tag chips disappear when their toggle is off while the local branch chip stays.
- [x] Manual: ran the app against a scratch repo with four stashes (one `stash -u`), two tags and three remote branches, one of them holding a commit no local branch reaches. Verified each toggle, and that the four stashes draw as four sibling lanes into their base commit.

### DIFF
--- a/src-tauri/src/modules/git/mod.rs
+++ b/src-tauri/src/modules/git/mod.rs
@@ -102,7 +102,7 @@ pub enum CommitOrder {
 /// `deny_unknown_fields` 是刻意的：搭配 `default`，一個拼錯或過時的欄位名
 /// （例如改名前的 `branch`）會安靜地取預設值，篩選就變成「全部分支」而且不
 /// 報錯。寧可讓它在反序列化階段就失敗、把錯誤送回前端。
-#[derive(Debug, Clone, Default, Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase", default, deny_unknown_fields)]
 pub struct GraphOptions {
     /// 篩選用的分支清單；空清單代表「全部分支」。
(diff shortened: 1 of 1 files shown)


mukiwu (approved):
已重新審查：stash ref 現在會跟隨 include_stashes 開關過濾，並補上回歸測試；remote、tag、stash 三組 toggle 的行為一致，CI 通過。

PR #409, opened 2026-09-07
feat(git): list the refs worth comparing against, and which one to suggest

yw-chan:
Part of #398, the first of the three backend pieces in the order laid out on that issue. Independent of #410 and #411 — all three branch from `master` and merge in any order.

## Problem

The comparison base selector needs to know what a repo is worth being compared against, and nothing here answers that. `branches()` comes closest, but it deliberately drops `*/HEAD` — correct where it is used, since a symbolic ref is noise on a ref chip, and exactly wrong here, because the pointer to a remote's default branch is the answer.

Deciding it in the frontend would mean the fallback order lives there, and the same ordering would exist twice as soon as anything else needed it.

## Changes

`git_comparison_bases` returns the candidates and the suggestion together, from three sources in the order they are preferred:

- **Each remote's `<remote>/HEAD`.** Every remote, not just `origin`: under a fork `origin` is your own copy and `upstream` is where the pull request goes, and both get compared against. A remote whose HEAD is unresolved is skipped rather than guessed at — `<remote>/main` is wrong on a repo whose default is anything else, and that ref genuinely may not exist (`git remote add` does not create it; `git remote set-head -a` does).
- **The current branch's upstream**, which answers a different question from the mainline: what is committed but not pushed.
- **A local `main`/`master`/`develop`**, for a repo with no remote at all.

`suggested` is the first of them, or null. A branch called neither main nor master, in a repo with nothing to fetch from, has no sensible mainline — offering an unrelated branch is worse than asking.

The function sits after `branches()` and its helper, which is where the comment about deliberately not reusing that function's `*/HEAD` filter can actually point at what it means.

Nothing else changes: one new command, registered, and no existing behaviour touched.

## Test plan

Four Rust tests, each building a real repository:

- two remotes, both defaults offered, and the suggestion one of them
- a remote whose `HEAD` was deleted, skipped in favour of the local `main`
- a tracking branch listed in its own right — and, when a branch tracks the mainline directly, that ref listed once rather than twice
- a repo with no remote and a branch called neither main nor master, yielding no candidates and no suggestion

`cargo test comparison_bases` — 4 passed. No frontend change, so the vitest suite is untouched.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -28,7 +28,7 @@ use modules::fs::{
 };
 use modules::git::{
     git_branch_checkout, git_branch_checkout_track, git_branch_create_at, git_branch_delete,
-    git_branches, git_cherry_pick, git_commit, git_commit_details, git_commit_file_diff,
+    git_branches, git_cherry_pick, git_comparison_bases, git_commit, git_commit_details, git_commit_file_diff,
     git_commit_range_file_diff, git_commit_range_files, git_commits_in_range, git_diff, git_fetch,
     git_file_at_rev, git_graph_log, git_log, git_merge, git_pull, git_push, git_push_delete,
     git_rebase, git_reset, git_resolve_repo, git_restore_file, git_revert, git_stage, git_status,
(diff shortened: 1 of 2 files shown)


mukiwu (approved):
審查通過：comparison base 候選與 suggested 由 backend 統一產生，符合 #398 的設計，沒有新增依賴或改動 TempoTerm 核心 workspace / pane 模型。與 #410、#411 視為同一功能組處理。

PR #410, opened 2026-09-07
feat(git): let a range be read from where the two sides parted

yw-chan:
Part of #398, the second of the three backend pieces. Independent of #409 and #411 — all three branch from `master` and merge in any order; I checked each pair with `git merge-tree`.

## Problem

`commit_range_files` and `commit_range_file_diff` are both two-dot: the literal difference between two trees. That is the wrong question for "what does this branch change" — every commit the base gained since the branch left it counts as the branch's own work, and files it never touched show up in the list, backwards, as deletions.

## Changes

Both commands take `merge_base`, and both build their argv through one `range_spec` helper.

The helper is the point rather than two format strings. These are two commands running two git invocations, and if only one of them learned about three-dot, the file list and the per-file diffs would disagree by exactly the commits made on the base since the branch left it — silently, with no error, just a file too many or too few. You called that out on the issue and it is the failure worth designing against.

Absent, the flag means two-dot, so every existing caller keeps the comparison it already asked for and no frontend change is needed. The default lives in the command layer; the library functions take a plain `bool` and say what they mean.

## Test plan

`three_dot_ignores_what_the_base_gained_after_the_branch_left_it`, next to the two-dot test it is the counterpart to, builds a repo where `main` and `feature` both move after diverging, then checks the two answers differ in the file only `main` touched — in the file list *and* in the per-file diff, which is the pair that has to agree.

`cargo test -- three_dot commit_range` — 2 pass; the existing range test still passes on the two-dot default. `tsc` clean, vitest untouched.

### DIFF
--- a/src-tauri/src/modules/git/mod.rs
+++ b/src-tauri/src/modules/git/mod.rs
@@ -1653,12 +1653,33 @@ pub fn commit_file_diff(repo_path: &str, commit: &str, file: &str) -> Result<Str
     Ok(diff)
 }
 
+/// The two ends of a range as git argv, either two-dot or three-dot.
+///
+/// Two-dot is the literal difference between the two trees. Three-dot starts
+/// from where they diverged, which is what "what does this branch change"
+/// means -- without it, every commit the other side gained in the meantime is
+/// counted as yours.
+///
+/// It is one helper because the file list and the per-file diff have to agree.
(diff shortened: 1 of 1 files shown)


mukiwu (approved):
審查通過：file list 與 file diff 共用 three-dot merge-base range 語意，保留既有預設行為，沒有發現阻擋合併的問題。

PR #411, opened 2026-09-07
feat(git): read a file at any rev, and say what makes one valid

yw-chan:
Part of #398, the last of the three backend pieces. Independent of #409 and #410 — all three branch from `master` and merge in any order; I checked each pair with `git merge-tree`.

## Problem

`file_at_rev` accepted `HEAD` and `:` and refused everything else — in both layers, as you noted on the issue. The Rust guard returned an error, and the TypeScript signature typed the parameter as those two literals, so nothing else could be passed in the first place. A comparison base is a branch, a tag or a hash, so both had to open up.

## Changes

The interesting part is what replaces the whitelist, which was doing two jobs at once: it picked the supported revs and, as a side effect, validated them — a literal cannot be anything else. Taking it away means the validation has to be said out loud instead of left implied.

So the rev is resolved before it reaches any argv:

```rust
run_git(repo_path, &["rev-parse", "--verify", "--quiet", &format!("{rev}^{{commit}}")])
    .map_err(|_| format!("unknown rev: {rev}"))?;
```

`--verify` makes rev-parse fail rather than echo the string back, and `^{commit}` refuses a tree or a blob: what is wanted is a point in history, not any object that happens to be nameable. `ensure_not_flag` stays and now covers the rev too, but it only catches a leading dash — which is why it was never enough on its own.

`HEAD` and `:` keep their shortcut: `:` is not a rev-parse rev at all, it is already the separator in `:path`.

A file missing at that rev is still an empty document rather than an error, which is what makes a new file read as all-added.

The second commit widens that last part. Reading the index side of an untracked file can fail with

```
fatal: ambiguous argument ':<path>': unknown revision or path not in the working tree
```

rather than the missing-file wording the function already recognised. Which of the two git picks varies by repository — the same shape of untracked file gives one message in a fresh test repo and the other in a real one — and the caller cannot tell them apart, so both now read as an empty document. It came from a diff surface showing "could not load the comparison" over a file it had itself just listed.

The TypeScript doc comment above the signature was stale too — it still named the two revs — so it goes with the signature.

## Test plan

`file_at_rev_reads_any_rev_and_refuses_one_that_is_not`, alongside the three existing `file_at_rev` tests, builds a repo with a tag, a branch and a later commit and reads the file at each; checks a missing path still comes back empty; and checks that a name git cannot resolve and an option-shaped one are both refused.

`a_path_git_will_not_resolve_reads_as_empty` covers the second commit: an ignored path reproduces that wording reliably, and both `:` and `HEAD` come back empty rather than erroring.

`cargo test file_at_rev` — 5 pass, the three existing ones included. `tsc` clean; the diff and Source Control suites pass, and no frontend caller passes anything but the two revs it always did.

### DIFF
--- a/src-tauri/src/modules/git/mod.rs
+++ b/src-tauri/src/modules/git/mod.rs
@@ -1138,13 +1138,28 @@ pub fn diff(repo_path: &str, staged: bool) -> Result<String, String> {
     }
 }
 
-/// Content of `path` at `rev`, where rev is limited to "HEAD" (last commit)
-/// or ":" (the index) — the only two versions the diff tab compares against.
-/// A file missing at that rev is an empty document, not an error, so new
-/// files diff as all-added.
+/// Content of `path` at `rev`. "HEAD" is the last commit and ":" is the index;
+/// anything else is resolved as a rev, so a branch, tag or hash can be read
+/// too — the comparison base in #398 needs that.
+///
+/// The two-value whitelist that used to stand here was doing double duty: it
+/// picked the supported revs *and*, as a side effect, validated them, since a
+/// literal cannot be anything else. Opening it up means the validation has to
+/// be said out loud, so the rev is resolved with `rev-parse --verify` before it
+/// is used. That both rejects nonsense and keeps a value that happens to look
+/// like an option out of the argv (`ensure_not_flag` only catches a leading
+/// dash).
+///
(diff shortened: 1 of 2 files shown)


mukiwu (approved):
審查通過：任意 revision 先經 commit 驗證，再進入讀檔流程；branch、tag、hash 與缺檔情況都有防護，對安全性是正向增強。

PR #414, opened 2026-09-09
feat(source-control): make the panel the all-changes page's table of contents

yw-chan:
Part of #397, the second of the three pieces, on top of #406 now that it has landed.

## Problem

#406 put every uncommitted change on one page, and left the Source Control panel beside it doing nothing useful: clicking a row opened a separate diff tab of its own, so the two surfaces showed the same files and disagreed about which one you were reading. A long page also gave no way back to a file except scrolling for it.

## Changes

While the all-changes page is the pane in front, the panel becomes its index rather than a second way in.

- **Clicking a row scrolls the page** to that file instead of opening a tab of its own.
- **The row the page is on is marked**, and it follows the scroll — including the first file while the page still sits at its top, and the enclosing folder when the file itself is collapsed out of sight in the tree view.
- **The panel's refresh button rescans both**, since while that page has the pane it is the only refresh control on screen.
- **The flat list is ordered the way the tree draws it**, so the panel and the page are one index of one comparison rather than two differently sorted lists of the same files.

The marking is read per row rather than by handing every row the answer, so a page scroll re-renders one row instead of the whole list.

## Test plan

`src/modules/diff/allChangesPanelLink.test.tsx` covers the pairing from both sides: a click scrolls rather than opening a tab, the mark follows the page, and the folder takes the mark when the file inside it is shut away. The Source Control suite covers the ordering and the shared refresh.

`tsc` clean. Full suite: 2170 pass, with two `PortsPanelView` tests failing only under a loaded parallel run and passing 16/16 in isolation — they fail the same way on master.

### DIFF
--- a/src/modules/diff/lib/allChangesLinkStore.ts
+++ b/src/modules/diff/lib/allChangesLinkStore.ts
@@ -1,0 +1,36 @@
+import { create } from "zustand";
+
+/** The file a click on a Source Control row is asking to be scrolled to. */
+export interface AllChangesFile {
+  /** Repo-relative path, as `gitStatus` reports it. */
+  rel: string;
+  /** Which of the two sections the row came from. */
+  staged: boolean;
+}
+
+interface AllChangesLinkState {
+  file: AllChangesFile | null;
+  /** Ask the all-changes page to scroll to this file once it can. */
+  request: (file: AllChangesFile) => void;
+  /** Read and clear the pending file; null if nothing is pending. */
+  consume: () => AllChangesFile | null;
+}
+
+/**
+ * The link between the Source Control panel and the all-changes page. The page
+ * is a singleton per space (openAllChangesTab always focuses the one tab), so
+ * there is no per-tab PaneContent field to route this through; the Git Graph
+ * tab's pending commit selection has the same shape and the same reason.
+ *
+ * Clearing on consume is what lets the same row be clicked twice: the second
+ * request is a real state change because the first left the field null.
+ */
+export const useAllChangesLinkStore = create<AllChangesLinkState>((set, get) => ({
+  file: null,
+  request: (file) => set({ file }),
+  consume: () => {
+    const file = get().file;
+    set({ file: null });
+    return file;
+  },
+}));
--- a/src/modules/diff/lib/allChangesLinkStore.ts
+++ b/src/modules/diff/lib/allChangesLinkStore.ts
@@ -9,60 +9,95 @@ export interface AllChangesFile {
 }
 
 interface AllChangesLinkState {
-  /** Panel to page: scroll to this file. A one-shot request. */
-  file: AllChangesFile | null;
-  /** Ask the all-changes page to scroll to this file once it can. */
-  request: (file: AllChangesFile) => void;
-  /** Read and clear the pending file; null if nothing is pending. */
-  consume: () => AllChangesFile | null;
+  /** Panel to page, by pane: scroll to this file. A one-shot request. */
+  file: Readonly<Record<string, AllChangesFile>>;
+  /** Ask one page to scroll to this file once it can. */
+  request: (pane: string, file: AllChangesFile) => void;
+  /** Read and clear that page's pending file; null if nothing is pending. */
+  consume: (pane: string) => AllChangesFile | null;
 
   /**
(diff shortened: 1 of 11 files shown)


src/modules/diff/lib/allChangesLinkStore.ts
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * The link between the Source Control panel and the all-changes page. The page
+ * is a singleton per space (openAllChangesTab always focuses the one tab), so
+ * there is no per-tab PaneContent field to route this through; the Git Graph
+ * tab's pending commit selection has the same shape and the same reason.
+ *
+ * Clearing on consume is what lets the same row be clicked twice: the second
+ * request is a real state change because the first left the field null.
+ */
+export const useAllChangesLinkStore = create<AllChangesLinkState>((set, get) => ({
```
mukiwu, line 55: [P2] Scope this store to the active All Changes pane. PaneTabContent can render multiple AllChangesTabContent instances after a split, but every instance writes to this global showing value and cleanup calls setShowing(null). Closing or switching one pane can therefore clear or overwrite the mark belonging to the other pane. Please scope the request/position state by leaf/pane, or make only the focused pane publish it, and add a two-pane regression test.

mukiwu (changes requested):
整體方向符合 #397：All Changes 與 Source Control 的導覽關係更清楚，refresh、folder mark、flat order 與右鍵 Show Diff 也都有補測試。

這次先 request changes，原因是 split pane 下的同步狀態仍可能互相污染：目前 link store 是全域 singleton，但已掛載的 pane 可以同時存在多個 AllChangesTabContent。請先將 file/showing/rescan 綁定到 leaf/pane，或確保只有 focused pane 能發布 showing；同時補上兩個 All Changes pane 同時存在、其中一個 unmount 的 regression test。

完成後請重新跑 targeted tests 與完整 CI；這個問題修正後，我會傾向接受並合併。

yw-chan:
已修正,請再看一次。

**做法**:把 link store 從全域單一值改成**按 pane 分開**(`file` / `showing` / `rescan` 都是 `Record<paneId, …>`),`AllChangesTabContent` 從 `PaneTabContent` 收到自己的 `paneId`,面板則讀「目前在前面的那個 all-changes pane」的條目——新增 `activeAllChangesPane()` 回傳 leaf id,原本的 `allChangesPaneActive()` 改成問同一個答案是否為 null。卸載時只刪自己那幾筆(`forget(pane)`),不再把還開著的那個 pane 的標記一起清掉。

我選按 pane 綁而不是「只有 focused 才發布」,因為後者只解決互相覆蓋,關閉時清掉別人標記的那半仍在;而且焦點切換的瞬間會有標記空窗。

**過程中抓到一個相關的 bug**:面板的 refresh callback 是 memoise 在 repo 上的,卻閉包了「目前在前面的 pane」。那個值在面板第一次渲染時是 null,之後永遠不會更新,所以重新整理其實從來沒推給頁面過——現在改成按下的當下才讀。

**測試**:新增兩個 pane 同時掛載、其中一個 unmount 的回歸測試(`allChangesPanelLink.test.tsx`),驗證各自的標記不互相覆蓋、請求只到指定的那個 pane、關掉一個之後另一個的標記還在。把 `forget` 改成清空全部可以讓最後那條斷言變紅,確認它咬得住。

`tsc` 乾淨;`src/modules/diff`、`src/modules/source-control`、`src/stores` 共 351 支全過;完整跑 2187 過,兩支紅的是 `PortsPanelView Apple Intelligence` 和 `FileFinder keyboard navigation`——同樣設定連跑兩次會一次紅一次綠,在 master 上也是這樣。

mukiwu (approved):
## Standards

Approve。最新 revision 已把 All Changes link store 與 rescan/showing 狀態改成按 pane 隔離，並補上雙 pane 與 unmount regression test，解除了先前的 split-pane 狀態污染。`tsc --noEmit`、diff check、4 個 targeted test files 共 184 tests，以及 GitHub CI 都通過。

## Spec

符合 issue #397：Source Control panel 作為 All Changes 的目錄、檔案標記、folder mark、refresh 與導覽行為都維持既有產品方向；這是聚焦在 Source Control / All Changes UX 的功能增強，沒有改動 Tempo Term 的核心精神或大幅擴張套件。建議接受並合併。

PR #424, opened 2026-09-11
feat(diff): choose what the all-changes page compares against

yw-chan:
Part of #398.

## Problem

The all-changes page from #397 reads one thing: what is uncommitted right now. That is the question you ask before committing, and the only one it could answer. Reviewing what a branch changed against `origin/main`, or reading a commit's changes on a whole tab instead of in the details panel's few hundred pixels, meant opening files one at a time.

## Changes

**A base the page compares against.** A control in the page header, per repository and only for as long as the app runs — a base chosen three days ago is more confusing than no base at all, and a ref can be rebased away inside one session. It offers the refs worth comparing against (the remote default branch, the tracking branch, then branches and tags by recency), takes a pasted hash, and takes a range typed as `a..b`. Uncommitted work folds in or not.

A branch is compared from where it and HEAD parted, so commits the mainline gained after you left are not counted as yours. A range is compared literally, two-dot: neither end is on disk.

**Four commands behind it.** `comparison_bases` lists the refs worth offering and which one is which; `tags` the same for tags; `resolve_rev` answers whether git knows a ref, so a pasted hash is checked before the page moves rather than after; `diff_from_base` reads the base against the working tree in one pass rather than stitching committed and uncommitted diffs together — a reviewer judges the total, not two halves.

**The graph's details panel can hand a selection over.** One commit, two commits, or the working-tree row #415 added: each opens the page on what it is showing. Both ends of a commit's range are named outright rather than one of them written `X^`, since every caller already holds the parent and `47c9cee^..47c9cee` is two near-identical strings where two hashes say at a glance what is compared.

**What a base cannot answer.** A ref that has gone since it was chosen, a file that exists only on disk, a rename reported as a pair of paths rather than as a whole file added — each is shown as what it is rather than as an error.

## Out of scope

#398 also asks for right-click entry points in Git Graph — "compare everything against the current branch" on a branch, and "open this commit's changes in a tab" on a commit. Those are a separate PR, following this one: a context menu is a new surface, and arriving through one has to set the tab's base outright, including when the singleton tab is reused — its own behaviour to test. Raised in #398.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 2322 passing
- [x] `cargo test --lib modules::git::` — 96 passing. One pre-existing failure, `restore_file_reverts_unstaged_change`, which fails on master too: it writes `"original\n"` and reads back CRLF on Windows
[...]

PR #430, opened 2026-09-14
feat(diff): name the code a collapsed stretch leads into

yw-chan:
## Problem

A collapsed stretch in a diff says how many lines it is hiding and nothing else. Scrolling a long file, every bar reads the same, and neither the code above a change nor the code below it says which function you are in — the thing `git diff` puts after the `@@`, and what GitHub and VS Code both show.

`@codemirror/merge`'s `collapseUnchanged` builds that bar itself and writes its own text (`CollapseWidget.toDOM`, `state.phrase("$ unchanged lines", lines)`). There is no hook for the label, for the widget, or for what pressing it does, so the bars here are built rather than configured.

## Changes

**The bar names what it leads into.** git's own heuristic — the nearest line at or above the change that starts in the first column with a letter, `_` or `$` — so it is wrong where git is wrong and right where git is right. The last stretch in a file gets no name: there is no change below it to be inside anything.

**Openable in steps.** Twenty lines up, twenty down, or the lot. A press never leaves a sliver: a stretch of twenty-three opens whole rather than hiding three lines behind a bar that costs a row of its own to say so, and the arrow's label promises the number it will really open. An arrow with nowhere to grow — the first thing in the file, or the last — is drawn dead rather than left to do nothing.

**The way back is where it already was.** A stretch opened all the way leaves no bar to press, so its fold control sits in the gutter beside the line it starts at, which is where #393 put it.

**Both halves move together.** A split opens the same stretch on both sides, paired by how many changes sit above it: the two documents are different lengths, so neither an offset nor a place in a list will do.

**Keyboard, and the reader's own settings.** The arrows answer Enter and Space; the bar's type follows the diff font setting; hover matches the fold gutter beside it.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 2304 passing. The one failure is `PortsPanelView`, which this branch does not touch: it waits a second for markdown to render, which a loaded machine running the files side by side does not always manage. Five runs of that file on its own all pass
- [x] Every stretch of three hundred generated diffs, on both sides, against a copy of the library's `buildCollapsedRanges`. The library does not export that function, and a copy is what says so when the two answers part — a chunk covering no lines on one side, which every pure insertion and deletion produces, is where they can. Both mapping bugs found while building this are caught by it, and by the three named cases beside it
- [x] Manually, against a repository built for it — a file whose head is a pure insertion, a file with three functions changed one line each, and a six-thousand-line lock file: the two columns line up across every bar; each bar names what `git diff` names after its own `@@`; a press opens the same stretch on both halves and the name follows what the bar now leads into; tab reaches an arrow and Enter presses it, repeatedly; the gutter acts only where it draws an icon; the bar's type follows the diff font setting; inline reads the same as split; and the lock file scrolls without a stutter

Opened as a draft behind #424, which is waiting on the two questions raised in #398. The two branches are independent — `merge-tree` is clean between them and either can land first — but there is no sense asking for a second review while the first is still in the air.


mukiwu:
#424 已經合併了，這支不用再等。

順帶說明那邊的結果：面板維持並存、自建 combobox 的例外接受了，合併前的三個小修在合併之前完成，剩下的收尾追在 #440。

你準備好就把這支轉成正式的，我會排 review。

mukiwu (approved):
深審過了，折疊的算術我們沒找到錯的地方。拿 `@codemirror/merge` 自己的 `buildCollapsedRanges` 對了 200 組隨機 diff，兩側每根 bar 藏的行數完全一致；再用 120 組各按 8 次（含跨側配對），沒有 desync、沒有殘留碎片。邊界也都驗過：整檔無變更、整檔都是變更、剛好 5 行的區段、`doc.line()` 的範圍。植入 10 個突變抓到 8 個，所以「削 sweep 預算」那次 commit 沒有把覆蓋掏空。刪掉的 78 加 92 行沒有漏掉行為。

先合，下面幾條我們自己接手補，不用你再改一輪：

- `collapseRuns.ts` 的 `RunWidget` 沒覆寫 `estimatedHeight`，被它取代的函式庫 widget 回 27。CodeMirror 對還沒畫出來的 block widget 是拿 `lineHeight` 當估值，而這根 bar 比一行字高 3 到 4px，幾百根 bar 的 lockfile 總高會低估近千 px，捲軸滑塊會跟著跳
- `index.css` 的 `.cm-collapsedLines` 變成死碼了，這個 class 只有函式庫的 `collapseUnchanged` 會長出來，而 `diffViews.ts` 已經把它拿掉
- `diffViews.ts:42` 的註解還在講那個已經被移除的 `Compartment`
- 區段名稱尾巴會多一個空格，`text.trim().replace(/[{(:]\s*$/, "")` 只吃掉括號後面的空白

兩件不改、說明理由：

**bar 本體不能點是你刻意的設計，維持原樣。** 「全開走 gutter 圖示，一個動作一個地方」這個取捨註解裡寫得很清楚。副作用是兩份內容完全相同的檔案會得到一根 `atStart` 與 `atEnd` 同時成立的 bar，兩個箭頭都死，只剩 gutter 圖示能開。那個狀態在這個 App 裡幾乎碰不到，不值得為它破壞你定下的規則。

**箭頭的 `up` 是以下方那個變更為基準在講方向**，跟 GitHub 的 expand up 慣例一致，圖示也對得上，維持現在的用詞。

謝謝這支，把函式庫沒開的口子自己接起來，論證和測試都做得很紮實。

PR #431, opened 2026-09-16
fix(diff): hold a header where it was put while the all-changes page settles

yw-chan:
## Problem

Shutting a file in the all-changes page put its header back where it sat — once, in the frame the file closed. A file shut from half way down a long diff could still end up a screen away, with files the reader had already read sitting above it. A file clicked in the Source Control panel had the same problem: it landed, then drifted off the top.

## Root cause

The correction ran in a single layout effect, and the page is still settling at that point. Closing a long file makes the page short enough to pull its neighbours into the mount window, and each of them replaces the height it was holding with the height its editors actually take — a git read and a CodeMirror build later. Every one of those above the header moves it, all of them after the correction has already run. `scrollToSection` had no follow-up at all.

Separately, the jump that waits for editors (the landing on the first change) could fire *after* a panel row was clicked, taking the page off the header it had just been asked for.

## Changes

- The header is held rather than placed: the correction is re-applied on every later move of the page — a neighbour mounting or unmounting, any section reporting a new height — until it holds or its 2s window runs out. A scroll position the page did not write ends it immediately, and a jump supersedes it.
- `scrollToSection`, the panel's row click, takes the same hold instead of a single scroll.
- A pending jump is dropped when something newer holds the page. Read off a ref rather than state: both are set from effects, so the finisher still runs in that commit carrying the older instruction.

Deliberately not changed: the page cannot scroll past its own end, so a file with less than a pane's worth of content under it settles as high as the list goes. Holding empty space below the last file would buy those few headers the top of the pane at the price of a blank screen at the end of every list.

## Test plan

- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm exec vitest run` — 243 files pass
- [x] Two new tests in `AllChangesTabContent.test.tsx` fail against the code they fix: a header pushed down by a neighbour that measured late, and a panel-requested file losing the top while it loads
- [x] A third guards the reader taking the page back mid-hold


mukiwu (dismissed):
方向對，骨架我認可：把「修正一次」換成「hold 到穩定或 2 秒到期」是正確的形狀，沒加依賴、沒加 timer、沒加 listener，unmount 面乾淨，target 不可達時會收斂不會無限迴圈。兩個新測試我逐行推過，對未修的 code 確實會紅。

有一條想請你先修再合。

## `AllChangesTabContent.tsx:360` — `applyPin` 無條件覆寫 `pin.at`

`applyPin` 最後那行不分有沒有真的寫入，都把 `pin.at` 記成當下的 `scrollTop`。而釋放 pin 的判斷只有一處，在 `trackPosition` 的 rAF 裡（`:571-576`）比 `root.scrollTop` 和 `pin.at`。

所以只要有任何一次 `applyPin` 執行在「捲動位置已經變了」與「那個 rAF 跑到」之間，讀者那次捲動就被吸收進 `pin.at`，rAF 一比相等，pin 不會被放掉 —— 而 `applyPin` 在同一次執行裡已經先把頁面拉回 `pin.top` 了。

失敗情境：讀者關掉一個檔案後馬上往下捲兩個檔案繼續看；此時上方某個檔案的 editors 剛好建好，`DiffFileSection.tsx:282` 那個 `.then` 裡的 `onMeasure` 被呼叫（這條路徑是 promise resolve 觸發的，跟 frame 沒有對齊，不像 ResizeObserver 保證排在 rAF 之後）→ `applyPin` 把頁面拉回被關掉那個檔案的 header，`pin.at` 記成拉回後的值 → rAF 比對相等 → pin 存活 → 接下來 2 秒內每次 measure 都可能再拉一次。

「讀者優先」這件事現在是靠 task 交錯順序碰巧成立的，不是靠規則成立的。這正好是這支 PR 要消滅的那種症狀，只是換了個觸發點。

建議把釋放判斷搬進 `applyPin` 開頭：進來先比 `root.scrollTop` 和 `pin.at`，差超過容忍值就 `pinRef.current = null` 然後 return，讓兩個進入點共用同一條規則；`pin.at` 只在真的有寫入時才更新。

## 順手一起處理（小）

**`:345-352` 找不到 element 時 pin 永遠不會過期。** expiry 檢查排在 `!element` 的 early return 之後，element 一直找不到就一直不過期。而 `elementsRef` 是在 `[ordered]` 的 passive effect（`:277-288`）裡重建的，rescan 之後到 passive effect 跑完之間有一段空窗，脫離 DOM 的節點 `getBoundingClientRect()` 全回 0，`now` 會變成 pane 在視窗裡的 top 偏移（實際 40px 以上），照這個假值把頁面往上捲。情境：關掉檔案 X 之後 300ms 切回視窗觸發 focus rescan，而 X 已經在別的終端機被 stage 掉、從 status 消失。建議 expiry 檢查提到 element 檢查之前，並加 `element.isConnected`，不連通就一併清掉。

**commit message 的描述比程式做的多一點。** 「a pending jump is dropped when something newer holds the page」只處理了一個方向。反方向是：使用者在掃描還沒回來之前就點了 Source Control 某一列，等 `ordered` 落地的那個 commit 裡，requested effect（`:699`）先 `scrollToSection` 設好 pin，接著 landed effect（`:740`）的 `jumpTo`（`:639` 無條件清 pin）把它蓋掉，使用者點的檔案就被放掉了。這跟 master 行為相同、不是這支造成的迴歸，我不擋這個 PR，但 message 的描述會讓人以為已經涵蓋。要修的話 landed effect 開頭加 `if (pinRef.current) { landed.current = true; return; }` 就好。

## 測試

兩個新測試對未修的 code 會紅，這點成立；第三個是護欄，也確實鎖得住釋放邏輯。缺口是 **2 秒 window 的到期完全沒有被測到** —— 那是整個 hold 唯一的界線，希望補一個。另外 `PIN_SETTLE_MS = 2000` 用的是真實時鐘沒有 fake timers，CI 塞住時從 `fireEvent.click` 到手動 fire observer 之間超過 2 秒並非不可能，屆時會以看不出原因的 timeout 失敗。

`vi.unstubAllGlobals()` 放在 `beforeEach`，stub 會活到下一個測試開始才撤掉，中間包含 RTL 的 `afterEach` cleanup。目前沒有實際危害，但慣例上該放 `afterEach`。

mukiwu:
已合併，謝謝。

`applyPin` 無條件覆寫 `pin.at` 那條我們這邊接手修，順帶會把 expiry 檢查提到 element 檢查之前並加 `isConnected`。2 秒 window 到期的測試也會一起補。

PR #432, opened 2026-09-16
feat(git-graph): page history with skip instead of re-reading the window

yw-chan:
## Problem

"Load more" grew the window instead of moving it: 200 commits, then 400, then
600, always read from the top. A press cost as much as every press before it
put together, and the whole list was re-parsed, re-sent over IPC and laid out
again each time.

Worse, it walked into `graph_log`'s cap on a single call:

```rust
let limit = limit.clamp(1, 2000);
...
let has_more = commits.len() > limit;   // against the clamped limit
```

Past 2000 the request was silently cut back to 2000, and `has_more` — compared
against the value it was cut back to — stayed true for ever. So the button
remained on screen, reporting more history, while the backend was structurally
unable to return any. 2000 commits was a hard ceiling on what the graph could
ever show, and pressing the button did nothing at all.

## Root cause

The `skip` parameter to page with has been there since the graph was first
ported (ea6d884): `git_graph_log` takes it, the bridge takes it, `graph_log`
passes it straight to `git log --skip=`. Nothing ever filled it in. The one
caller grew `limit` instead, so every press re-read from row 0.

That is also how a cap on one call came to act as a limit on history. Nothing
decided 2000 was as far back as anyone may look; it is what happens when the
only way to ask for older commits is to ask for more commits.

## Changes

**A page is a page.** `loadMore` asks for `PAGE_SIZE` plus a 20-commit overlap,
skipping past what is already loaded, and appends the result. The cost of a
press no longer depends on how many presses came before it, and the cap is back
to capping one call.

**The overlap is what makes appending safe.** A page starts where the last one
ended, counted in rows of the walk, and the walk moves: a commit landing in
another terminal pushes every later row down one; a branch going away pushes
the other way. The page is joined to the list by finding the last loaded
commit's hash inside the overlap — so the seam is always a commit that exists
in both walks. Drift wider than the overlap cannot be stitched and falls back
to re-reading the window. The graph never gets a hole it cannot see.

**A page that was in flight when a whole-window reload landed is discarded.**
The list it measured itself against is gone; splicing onto the new one would
join rows from two different walks.

**Branches, worktrees and the working tree are no longer re-read on a page.**
None of them change because the reader scrolled further back.

**The pending-selection retry still re-reads the whole window.** It is hunting
for a commit as likely to be newer than everything loaded as older, and
appending only ever reaches further back.

The clamp keeps its 2000 and gains the comment it never had, saying what it is
actually for.

## What this costs

Two things behave differently, both on purpose.

**Past 2000 loaded, a whole-window reload truncates back to 2000.** Refresh and
the context-menu git operations re-read however deep the reader had paged, in
one call — which is exactly what the cap cuts short. This could not happen
before, because the list could never exceed 2000 in the first place.

Reaching it takes ten deliberate presses of "Load more" and then a git
operation, and nothing is lost but the depth: the commits are still there, five
more presses away. Restoring the depth properly means either asking for all of
it in one call — the same cost the cap exists to bound — or re-reading only the
head and keeping the tail, which is not safe: a merge puts the branch's commits
wherever their dates fall, which can be thousands of rows below the head, so
the tail would be missing commits that the rows above it point at.

**Rows already loaded are no longer refreshed by paging.** Re-reading
everything used to quietly drop commits that had gone away. Appending does not,
so the top of the list is a snapshot until something reloads it. That is not a
hole — the seam check guarantees the list is a contiguous run of one walk — it
is staleness, and it clears on any refresh or git action.

Both have the same eventual answer: knowing cheaply whether the walk changed at
all (`git show-ref --head`, plus HEAD's symbolic target and the stash reflog,
neither of which show-ref covers). That is a separate piece of work and is not
attempted here.

## Test plan

- [x] `npx pnpm exec vitest run` — 236 passed in `src/modules/git-graph`
- [x] A page asks for the rows after the ones loaded (`skip`), not for a bigger
      window
- [x] The overlap's rows are the seam, not new history — appending a page that
      hands back two known commits leaves three rows, not five
- [x] A page whose seam is lost re-reads the whole window rather than appending
      a run from a different walk
- [x] A page still in flight when a reload lands is dropped
- [x] Manually against a 24k-commit repo: past the old 2000-commit ceiling, and
      a press at page 15 costs what a press at page 2 costs


mukiwu (dismissed):
方向我同意：`skip` 本來就接好了沒人用，把它真正用起來、順手拆掉 2000 那個假天花板，是對的做法，PR 內文把成本與取捨也交代得很清楚。但有一個問題要先修，這支才能合。

## Blocker：generation 檢查擋不住「reload 已經在飛才按 Load more」

`reload()` 在進函式第一行就 `loadGeneration.current += 1`，之後才 `await`；`loadMore()` 在 `GitGraphTabContent.tsx:333` 抓的是當下的值。所以只有「Load more 先開始、reload 後到」會被擋掉，反過來不會：

1. reload 開始（gen 0 → 1），await 中
2. 使用者按 Load more，抓到 gen = 1，`loaded` 還是舊清單，`skip = loaded.length - 20`
3. reload 回來，`setCommits(log.commits)` 整份換掉
4. page 回來，`:343` 比對 `1 !== 1` 為 false，沒被丟掉；`:349` 用舊清單的最後一筆算 seam，`:360` 卻接到新清單後面

具體情境：視窗 200 筆，別的終端機剛 commit 一筆，使用者按 Refresh、趁轉圈時又按 Load more。新視窗是 `[C, old0..old198]`，page 是照含 C 的 walk 從 `skip=180` 抓的，seam 落在 old199，接起來之後 old199 整筆消失。

漏一列不是少一列而已：`graphLayout.ts:165-171` 的 lane 要等 parent 出現才釋放，洞上面那列的 parent 永遠等不到，那條 lane 被永久佔住，下面每一列往右擠；超過 `maxLane` 之後又會壓回同一欄，於是畫出根本不存在的父子線。而 Load more 按鈕（`GitGraph.tsx:608-616`）在 reload 期間沒有 disabled，也沒有任何遮罩，按得下去。

master 上的舊寫法雖然笨，但兩個 reload 互相覆蓋頂多是「舊視窗贏」，每份清單自己都是完整的，不會拼出破洞，所以這是這支新增的失效模式。

修法一行就夠：`reload` 在 `await` 之後、`setCommits` 之前再 bump 一次 generation（catch 分支也要），兩種交錯順序就都擋得掉。不要改用 `commitsRef.current !== loaded` 比身分，React 18 的 auto-batching 之下 ref 可能還指著舊陣列。

## 其他（不擋合併，但希望一起處理）

**`mod.rs:1417-1421` 新加的註解與實際不符。** 註解說呼叫端不會接近 2000，但 `reload(repo, limit, ...)` 傳的是累積的 `limit`，按 11 次 Load more 之後就是 2200，任何右鍵 git 操作、Refresh、fetch 都會拿它去打然後被砍回 2000。這個行為本身是你列的取捨 #1，我接受（master 上根本過不了 2000，這是往好的方向走），但註解要寫實話。順帶一提 `:1409-1410` 的 doc comment 從頭到尾沒提 `skip`，而這支讓 `skip` 第一次變成關鍵參數。

**Rust 側沒有任何測試傳過非 0 的 `skip`。** `graph_log` 的十處測試呼叫全部傳 0。希望補一個「`skip=N` 那頁剛好接在 `skip=0` 那頁後面、`has_more` 正確」的測試，在真 repo 上比對 `graph_log(limit=10, skip=0)` 的尾段與 `graph_log(limit=5, skip=5)`。

**`GitGraphTabContent.test.tsx:232` 那個測試對舊程式碼也會過。** 第二次 mock 回傳的是「從頭全部」，舊的整窗重讀直接換掉清單，畫面一樣是三列、Load more 一樣消失。要有鑑別力，第二次應該只回「overlap ＋ 新的」，像上面那個測試的形狀。其餘三個測試我逐一對舊碼推過，都會紅，沒問題。

**Load more 在 page 飛行中沒有任何回饋。** `loadingMore.current` 擋得掉重複點擊，但按鈕既不 disabled 也不轉圈，大 repo 上跑一兩秒使用者只會覺得壞了而狂點。這同時也是上面那個 blocker 的入口，一起處理最省事。

**topo order 下的去重（低機率，便宜的加固）。** seam 只比對最後一筆的 hash，之後 `slice(seam + 1)` 全收。`--date-order` 下已載入的 commit 幾乎不可能被推到 seam 之後，但選 `topo` 時多一個 merge 可能重排 walk。`fresh` 再過一次 `new Set(loaded.map(c => c.hash))` 就好。

mukiwu:
已合併，謝謝。

generation 那一行（`reload` 要在 await 之後、`setCommits` 之前再 bump 一次）我們這邊接手修，並會補上反向交錯的測試。Rust 的 `skip != 0` 測試與那段和實際不符的 clamp 註解也會一起處理。

Issue #437, opened 2026-09-17
chore(git-graph): pin the lane ceiling with tests and drop the unreachable fade code


mukiwu:
Follow-up to #425, merged with these leftovers. No user-visible defect.

**The fade never reaches outside the gutter.** A visible stroke needs `x < fadeTo = paddingLeft + columns * laneWidth + 12`, while `gutter >= paddingLeft + columns * laneWidth + 24`, so `fadeTo` is always `gutter - 12`. At the default geometry the gutter is 240, `fadeTo` 228 and `rowIndent` 228: the fade lands exactly where the text starts. That makes `style={{ overflow: "visible" }}` at `GitGraph.tsx:468` dead, and the comment at `:157-172` describing lines running past the gutter over the rows misleading. The risk is later: trim `GUTTER_TRAIL` from 24 to 8 to save width and that harmless overflow starts drawing lines over commit messages, with no test to catch it.

**The ceiling is not pinned by a test.** `GitGraph.test.tsx:718-740` only asserts that no node has opacity 0 and that 40 lanes draws fewer nodes than 20. Mutating `graphLayout.ts:171` to `laneX(columns - (columns >= 20 ? 2 : 1), ...)`, which silently drops lane 19 in the wide band while the gutter still has room for it, leaves all 242 tests green. Add: at exactly `maxColumns` the node count equals the visible row count, and at 21 it is one short.

**"The row is a click target in its own right" has no test.** Both the commit message and the PR body use it to justify dropping the faded node. It holds today (clicking row `r39` at 40 lanes does reach `onSelectCommit`), but nothing stops a later change to the row from taking it away.

Smaller:

- `GitGraph.tsx:227-228, 585, 589` is dead. A node's x is `paddingLeft + lane * laneWidth + 12` and the fade window is exactly one lane wide, so no integer lane can land inside it: `nodeOpacity` only ever returns 1 or 0, `opacity: shown` is always 1, and the `z-0` branch is unreachable (and would be wrong, putting the node behind its row).
- `GitGraph.tsx:169` says "Three lanes later it is gone" while `LANE_FADE_LANES` is 1.
- `graphLayout.ts:124`: `maxColumns` is only meaningful at `maxLane + 1` or above, and nothing enforces it.
- The six glow numbers are written twice, in `GitGraph.tsx:215-220` and `index.css:647-653`, so changing the CSS default does not move the narrow-lane scaling.

[...]

PR #425, opened 2026-09-12
feat(git-graph): narrow the lanes instead of stacking them on one column

yw-chan:
## Problem

`laneX` collapsed every lane past `maxLane` (5) onto the last column, so a
graph wider than six lanes drew two branches at the same x — nodes of
different colours sitting on one line, and lines running down a lane that is
not their own.

It is not a hypothetical. Laid out over the ref set the graph actually walks
(`build_log_refs()` — `--branches --remotes --tags`), this repo reaches lane
index 7: **44 commit nodes land on x=98**, which lanes 5, 6 and 7 share.

| | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| rows at this width | 269 | 234 | 143 | 76 | 41 | 15 | 3 | 18 |

#407 hit the same collapse from the other side, where four stashes were
enough to draw a parent/child line between two siblings.

## Root cause

The gutter was a cap rather than a budget, and its width was worked out by
hand in three places that had to agree and did not have to: the track column,
the row indent, and the load-more button. The row indent was hardcoded at
112px for a widest node of 98px — so the moment the lanes could outgrow six
columns, the nodes were drawn on top of the commit hashes.

## Changes

**A ladder, not a cap.** Up to six lanes take the full 14px and nothing about
today's graph moves. Past that they narrow within the same 124px — 12px at
seven, 10px at eight — so the commit messages keep every pixel they have.
Only once the lanes cannot narrow further does the gutter grow.

| lanes | lane width | gutter |
| --- | --- | --- |
| ≤ 6 | 14px | 124px (unchanged) |
| 7 | 12px | 124px |
| 8 | 10px | 124px |
| 9–20 | 10px | 130 → 240px |
| 21+ | 10px | 240px, and the rest run outside it |

**One number, not three.** Which width applies is not known until every commit
has claimed a lane, so `computeGraphLayout` assigns the lanes first and
resolves x after. It returns the width and the gutter it used, and the three
callers take it from there instead of deriving it again.

**Past the ceiling, a soft edge rather than a stack.** Twenty columns is where
the gutter stops growing — a repo really can have fifty branches live at once,
and no gutter wide enough for them leaves room for the rows it is supposed to
annotate. The lanes past it are not stacked onto the last column, which is the
same collapse this PR set out to remove with the threshold moved. They keep
their real position, run outside the gutter, and a gradient takes them to
nothing within one lane's width.

[...]

PR #441, opened 2026-09-17
feat(git-graph): carry a line on past the last row it can reach

yw-chan:
## Problem

A commit whose parent is not in the loaded page gets no line at all. It hangs
there with nothing drawn below it, which reads as a branch that stops here
rather than one that carries on past what has been loaded.

It is not only the last row. `--date-order` can put a parent further down the
walk than the page reaches, so the gap appears in the middle of the graph: four
hundred commits of one repo to hand have two of them, and one is row 204.

## Changes

The line runs on to the edge of the last row's band.

Nothing has to be fetched to draw it. The commit names the parent, so it exists;
a walk is closed under its parents, so that parent is not somewhere else, it is
further down. Its lane goes on waiting for it and is therefore never freed or
reused — a line straight down that lane draws only what the lane bookkeeping
already says, and where the parent actually sits is not needed to say it.

**Where it stops matters.** Not the foot of the canvas: there is a whole row of
blank below the last node, the strip the load-more button sits in, and a line
running through it reads as escaping the graph rather than carrying on past it.
Not the last node's own height either — a lane with no node there would stop in
mid-air, level with someone else's dot. Half a row below the last node is the
edge of its band, and every line that leaves the page stops on that same edge,
so they read as one edge rather than as lines that each gave up somewhere.

**Only while there is more to load.** Once the walk is exhausted, a parent that
still will not resolve is a shallow clone's graft boundary — the history really
does stop there, and a line promising more below it would be a lie. The layout
reports the open ends either way and the caller, which is the one that knows,
decides whether to draw them.

**Two things had to be right for the line to land where it belongs.** It is
drawn on the lane reserved for that parent, not on its child's: a merge's second
parent claims a lane of its own, and drawing it on the child's would hide it
under the line to the first parent. And that lane has to count toward the
width — it carries a line but no node, so counting only nodes left it out, and
the column ceiling then clamped its line onto a lane it does not belong to,
which is the overlap #418 exists to avoid.

## Against #437

#437 reads the fade from #425 as unreachable, and this branch continues in that
direction, so the two were checked against each other.

They agree. An open end is `M x y L x pageEdge` — one vertical stroke at
`laneX(lane)` for an integer lane, so it takes exactly the x values the nodes
take. The fade window is one lane wide (`fadeFrom = laneX(columns - 1)`,
`fadeTo = laneX(columns)`), so no integer lane lands strictly inside it and
`nodeOpacity` still only ever returns 1 or 0. The stroke has no horizontal
extent either, so it cannot paint outside the gutter and does not depend on the
`overflow: visible` #437 calls dead.

Two edges are worth naming rather than leaving to be discovered:

- Past the column ceiling the gradient is fully transparent, so an open end out
  there paints nothing — the same answer #425 gives the node on that lane, which
  it leaves out. Consistent, and invisible either way, but it is paint spent on
  nothing if the ceiling is ever reached.
- A lane held open for an unreached parent now counts toward the width even
  though it carries no node. That is deliberate — without it the ceiling clamps
  the line onto a lane it does not belong to — but it means "widest" and "number
  of nodes" are no longer the same quantity, which the node-counting test #437
  proposes should be written knowing. A fixture whose parents all resolve has no
  open ends and is unaffected.

If #437's cleanup lands first, `strokeFor` is what these lines are painted with;
the conflict is textual, not behavioural.

## Test plan

- [x] `npx pnpm exec vitest run` — 268 passed in `src/modules/git-graph`
- [x] `computeGraphLayout` reports an open end for an unresolved parent, on that
      parent's lane, and none once the walk is exhausted
- [x] The lane an open end holds counts toward the width, so it is not clamped
      onto a neighbour
- [x] `GitGraph` draws the line only when there is more to load, and stops it
      half a row below the last node
- [x] Manually against a 24k-commit repo, where the first page has two of them


mukiwu (changes requested):
方向對，論證也站得住。我逐條查過你這條「保留的車道就證明 parent 在下面」會不會說謊，結論是不會：

- 分支篩選與 remotes／tags 開關只是換 `git log` 的 tip，而 `git log A B` 對 parent 封閉，不會漏掉某個 commit 的 parent
- stash 的輔助列在 Rust 端就把 helper 連同指向它的 parent 一起清掉了，前端拿到的 parent 不會懸空
- 工作區那一列不在 `commits` 裡，`pageEdge` 有把 `rowOffset` 算進去，位置正確
- #432 的分頁接縫：`--date-order` 和 `--topo-order` 都保證 parent 不會排在 child 之前，所以已載入的前綴對 parent 封閉；接縫掉了會退回整窗重讀，中間不會開洞

有一件事想請你合併前補上。

## merge 的那條線跟它的節點是斷開的

`graphLayout.ts:375-381` 的 `openEnds.push` 取的 `x` 是保留車道的 `laneX(link.lane)`，`y` 卻是子節點那一列的 y；`GitGraph.tsx:522` 再用 `M ${end.x} ${end.y} L ${end.x} ${pageEdge}` 畫一條純垂直線，中間沒有任何彎回去接節點。

拿這個 repo 的真實歷史切在第 400 列（也就是按一次載入更多之後的位置）量到的：

```
layouts m = { x: 28, y: 20 }
openEnds  = [{ lane: 1, x: 42, y: 20 }]   // dx = +14px
paths     = ["M 28 20 L 28 56", "M 42 20 L 42 74", ...]
```

節點在 x=28，線從 x=42 起筆，同一個 y，中間空著 14px。掃過 20 到 1000 之間每一個切點，981 個裡有 169 個至少出現一條這種線，大約 17%。切在 404 的時候那條線離底部還有兩列，等於一條約 90px 的線從 merge 節點旁邊的半空中開始 —— 正是你 commit message 自己寫要避免的 stop in mid-air。

修法：`OpenEnd` 多帶一個子節點的 x，前端在 `end.x !== childX` 時沿用 `edgePath` 既有的 merge-in 彎（`M cx cy C ... L x pageEdge`），第一列先彎進保留車道再直下，跟已經解析的 merge 畫法一致。大約十行。

## 兩個記著就好，不擋

**`graphLayout.ts:312-316` 的 `widest` 不知道 `hasMore`。** layout 無條件把「沒有節點但被 parent 佔住」的車道算進 `widest`，但要不要畫線是 `GitGraph.tsx:517` 用 `hasMore` 決定的。兩邊不同步時（shallow／graft 邊界，或接縫掉了還沒重讀完的那一瞬間）會留下一欄完全空白，而且 `lanes` 一變會連帶動 `laneWidth`、`gutter` 和 `fadeFrom`，整排車道橫移。這個 repo 的真實資料量不到影響，未解析的車道從沒超過有節點的最寬車道，所以是冷路徑。

**`hasMore` 擋不住走到一半的 graft 邊界。** shallow clone 的 graft 讓 `git log` 提早停，parent 名字還在但永遠不會被走到，這時線承諾的是一段永遠不會載進來的歷史。終端機 App 裡 shallow clone 很少，也沒有便宜的修法，我傾向留著，在註解裡承認這個縫就好。

## 測試

五個斷言裡四個對未修的 code 會紅，實測過。`draws nothing once the walk is exhausted` 舊 code 也是綠的，它是 regression guard 不是紅燈測試，這樣沒問題，只是要知道。

缺口是：沒有測試釘住線停在哪裡（`pageEdge`，以及工作區那列開著時的位移），也沒有測試釘住 merge 那條線有沒有接到節點 —— 也就是上面唯一真的畫錯的那個 case 正好沒被蓋到。補那個彎的時候一起補一個就好。

## 里程碑

這支會落在 v0.4.2。v0.4.1 今天就要凍結發版，不再收新的功能。

mukiwu (approved):
先合。merge 那條線跟節點斷開的問題我們自己接手修，跟著 #441 的 changelog 一起進來，不用你再改一輪。

修法就是原本說的那個：`OpenEnd` 多帶子節點的 x，`openEndPath` 在兩邊 x 不同時沿用 merge-in 的彎，第一列先從節點彎進保留車道再直下。有一點跟原本講的不一樣：parent 落在左邊時不能照抄 `edgePath` 延後彎的做法，因為保留車道是空的沒東西可蓋，反倒是子節點自己的車道已經在畫到第一個 parent 的線，延後彎會壓上去，所以兩個方向都走同一個彎。另外補了三個測試釘住線的起點與終點。

`widest` 不知道 `hasMore` 那條維持現狀，graft 的縫照說好的寫進註解了。謝謝這支，論證寫得很清楚。

PR #442, opened 2026-09-17
feat(git-graph): show the search where its matches are

yw-chan:
## Problem

Searching the graph said there were 140 matches and left the reader to find them
by pressing the arrows. Nothing on screen said which rows matched, or what in
them did — #426 made the graph keep every commit drawn while a search runs, so
the matches sat unremarked among everything else.

The counter had a second problem. It read the selected row, so:

- clicking any other commit walked off with the reader's place in the search;
- clicking one nothing matched read as **"0 / 140"** — a number that looks like
  "no matches" and means "you are not standing on one";
- closing the details panel cleared the selection, and with it the counter.

## Changes

**The rows mark what matched**, in whichever of the three searched fields
carried it: a query on an author's name marks the author, not a message that
happens to contain the same letters. The date is left alone because the search
does not read it — marking it would claim a match the counter does not count.
`commitMatches` comes out of `findCommitMatchIndexes` so the count and the marks
cannot disagree about what "matches" means, and so the list of searched fields
is written down once.

**The counter gets its own cursor**, keyed on the hash rather than an index, so
paging more history in or a reload reordering rows cannot quietly move it onto a
different commit.

| | cursor | selection (and the details panel) |
| --- | --- | --- |
| arrows / Enter | moves | moves — the panel still follows a search |
| clicking a row the search found | moves | moves |
| clicking a row it did not | **stays** | moves |
| closing the details panel | stays | cleared, as before |

A query starts on its first match rather than nowhere, so typing reads "1 / 140"
the moment it lands. Emptying the box ends that search — the next query starts at
its own first match — while editing one keeps the cursor as long as the commit it
names is still a match, so a trailing space does not send the reader back to the
top.

**Only the cursor's row is tinted.** Colouring every match leaves the row the
counter points at with nothing to tell it apart: a band of coloured rows, and the
reader hunting through it for where the next arrow press leaves from. Which rows
matched is said by the marks in their text, and the cursor's row marks its own
run more strongly, the way a find bar does.

## Two things that only show up on screen

Both live in `splitOnQuery` and are the reason it does more than call `indexOf`
in a loop.

**Matches that touch come back as one run.** Each mark is padded and pulled back
by the same pixel — inline padding alone would shove the rest of the line along,
and with several marks in a message the row's text visibly shifts as the reader
types. Two adjacent marks therefore overlap where they meet, and a translucent
background drawn twice there is a darker seam through the middle of a word: "off"
searched for "f".

**Nothing is marked when case-folding changes length.** Offsets found in the
lowercased string are used to cut the original, which holds until a character
folds to a different length. "İ" (U+0130) lowercases to two code units and is the
only character in the BMP that does, so every mark past one landed a character
late: "İsmail" searched for "mail" marked "ail", a one-character match marked
nothing inside an empty `<mark>`, and a cut could land inside a surrogate pair and
split an emoji across two elements. Marking nothing is better than marking the
wrong characters. The round-trip test cannot see this class of bug — joining the
runs back up still gives the original string; only the cut points are wrong — so
it has its own test that can.

## Costs

The first arrow press now goes to the *second* match, because the counter has
already put the reader on the first. That also means a search begun while reading
row 500 no longer jumps to the next match below them; it starts at the top. The
two cannot both hold: having just said "1 / 140", the next press has to be 2.

## Test plan

- [x] `npx pnpm exec vitest run` — 2403 passed (`PortsPanelView` fails
      intermittently on this machine under load, on master too)
- [x] `splitOnQuery`: every occurrence, case-insensitive, literal rather than a
      pattern, touching matches merged, folding that changes length marks
      nothing, and the runs always rebuild the original string
- [x] The marks land in the field that matched, and never on the date
- [x] The cursor survives a click on an unmatched row, moves onto a matched one,
      recovers when its commit stops matching or stops being loaded, and starts
      over when the box is emptied
- [x] The tint is on the cursor's row and does not follow a click — the one line
      wiring the two together, which the counter assertions alone left untested
- [x] Manually against a 24k-commit repo
- [x] Merged against #441 in both orders: clean, same resulting tree, and that
      tree typechecks with all 292 graph tests passing

## Scope

Four commits: the pure functions, the view and the cursor, a round of review
fixes kept separate so they can be read against what they changed, and one
that only moves a block of tests.

That last one is for #441's benefit. Both branches append a `describe` to the
end of `GitGraph.test.tsx`, which is a conflict at the last line with nothing
to decide — whichever lands first leaves the other to resolve it by hand.
Moving these cases up beside the compare-mode highlighting ones, which ask the
same question of what a row shows, puts the two additions in different places.
Merging the two in either order is then clean, and both orders produce the
same tree; `GitGraph.tsx` itself already merged on its own.

Nothing here searches git. A query still only reads the commits already loaded,
which is what "(loaded)" in the counter's tooltip has always meant, and pressing
"Load more" still fetches the next page of history rather than the next page
containing a match.


mukiwu (approved):
深審過了，正確性上沒有阻擋項。`splitOnQuery` 的 offset 數學我們抽出來對參考實作 fuzz 了 30 萬組（字母表含 `İ`、`ß`、`é`、emoji），四個不變式全過。那道長度保護也是精確的：列舉整個 0x10FFFF，`toLowerCase()` 會改變長度的 code point 只有 U+0130 一個，而且沒有任何一個會變短，所以 `haystack.length === text.length` 真的保證索引一對一，你的說法成立。沒有 XSS 面，query 只當 `indexOf` 的字面 needle，從不進 regex。#445 的 `loadGeneration` 兩次 bump 在合併後完好。

先合，下面幾條我們自己接手補：

**計數器說「1 / 140」，但畫面不會捲過去。** `GitGraph.tsx` 的捲動 effect 只吃 `activeHash`（selection），游標 fallback 設的 `cursorHash` 沒有任何 effect 會捲。而 `currentMatchIndex` fallback 成 0 之後，`(0 + 1) % len` 直接跳到第 2 筆，所以第一筆 match 按一次鍵到不了，要按 ↑ 繞一圈或自己捲回去。

失敗情境：讀者捲到第 500 列再開搜尋、輸入 `fix`，工具列顯示「1 / 140」，但它指的那一列在畫面上方 490 列處，連 tint 都看不到。這跟你這支要修掉的「0 / 140 看起來像沒找到」是同一類問題，計數器講的話畫面兌現不了。修法是加第二個只吃 `currentMatchHash` 的捲動 effect，不碰 selection，所以不會每按一次鍵就打一次 `gitCommitDetails`，那正是你刻意避開的。

- 「是當前 match 但沒被選取」那一列漏了 `hover:bg-bg-elevated/40`，滑過整排時只有那一列不亮
- 沒有測試釘住「query 命中 hash 時要標 hash 欄」。把 `<Marked>` 換回 `{commit.hash}`，整套 graph 測試照樣全綠，但 `commitMatches` 明明讀 hash，等於計數器算了一筆、列上卻沒有記號
- 游標對 `loadMore` 追加分頁的行為沒測。我們人工追過是正確的，補一條守著

有一件維持現狀：shift+click 建立比較範圍、以及 ArrowUp/ArrowDown 逐列瀏覽也會移動搜尋游標，因為都走 `onSelectCommit`。行為本身說得通，讀者確實站在那個 match 上，我們補一句註解寫下來就好，不改。

謝謝，游標改用 hash 而不是索引這個決定是對的。

